PR #1, opened 2019-10-25
Improve async interface

parasyte:
**[WIP]**; not ready to be merged.

Copy of publish example from the docs:

```rust
use json::object;

let pubnub = PubNub::new("demo", "demo");
let status = pubnub.publish("my-channel", object!{
    "username" => "JoeBob",
    "content" => "Hello, world!",
}).await;
```


src/pubnub.rs
```diff
@@ -0,0 +1,491 @@
...
+                self.subscribe_key.clone(),
+                channels,
+                ChannelMap::new(),
+            );
+
+            // Spawn the subscribe loop onto the Tokio runtime
+            tokio::spawn(subscribe_loop.run());
+
+            debug!("Waiting for long-poll...");
+            guard
```
pnehrer: Hmmm why is this necessary -- do we even need the Ready message?

src/pipe.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// Only sent from `SubscribeLoop` to `PubNub`, and only in unit tests.
+    #[cfg(test)]
+    Exit,
+}
+
+/// # Type of listener (a channel or a channel group)
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) enum ListenerType {
+    Channel(String), // Channel name
+    _Group(String),  // Channel Group name
```
pnehrer: Why not just `Group`?

src/pipe.rs
```diff
@@ -0,0 +1,54 @@
...
+/// See [`PipeMessage`] for more details.
+#[derive(Debug)]
+pub(crate) struct Pipe {
+    pub(crate) tx: PipeTx, // Send-side for bidirectional communication
+    pub(crate) rx: PipeRx, // Recv-side for bidirectional communication
+}
+
+/// # The kinds of messages allowed to be delivered over a `Pipe`
+#[derive(Debug)]
+pub(crate) enum PipeMessage {
```
pnehrer: Maybe we could use two kinds of Pipe -- client to SubscribeLoop and SubscribeLoop to client? That could address the "only sent from" constraints.

src/pipe.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// A stream for a channel or channel group is being created.
+    ///
+    /// Only sent from `PubNub` to `SubscribeLoop`.
+    Add(ListenerType, ChannelTx),
+
+    /// The `SubscribeLoop` is ready to receive messages over PubNub.
+    ///
+    /// Only sent from `SubscribeLoop` to `PubNub`.
+    Ready,
```
pnehrer: I'll continue reading the code but I'm not quite following why we need the Ready message and can't simply await and consume channel messages...

src/pipe.rs
parasyte: This is still unimplemented...

src/pipe.rs
parasyte: Yeah that thought crossed my mind, as well. Particularly as I was writing the doc comments for the PipeMessage variants.

src/pubnub.rs
parasyte: You're right, the `Ready` message is no longer needed. It was in place before the mutex was added. With that, the `mpsc` channel bound can also be set to 1. Will make the adjustments.

parasyte: oooh, I am completely mistaken. We can't remove the ready event for two reasons:

1. In the case that a `SubscribeLoop` already exists, this message will carry the `id` necessary for dropping the stream.
2. In the other case where a new `SubscribeLoop` needs to be created, this test (and others like it) will fail without waiting: https://github.com/stephenlb/pubnub-rust/blob/50081622f6e2caa496836ea8516bfb536508bd88/src/lib.rs#L127-L143

In the latter case, we can move the wait-for-ready signal to `#[cfg(test)]` at the cost of missing an opportunity to signal to the listener that the stream is actually capable of receiving messages immediately.

parasyte: Well, not _completely_ mistaken! We can still drop the `mpsc` bound to 1.

PR #2, opened 2019-11-08
Enable CI

parasyte:


parasyte:
Travis doesn't run the example, but it checks it with clippy! :D

PR #8, opened 2019-12-20
Decouple core implementation from hyper and tokio and refactor subscriptions

MOZGIII:
This PR splits the `pubnub` crate into a `pubnub-core` and `pubnub-hyper` crates, removing `tokio` and `hyper` dependencies from the `pubnub-core` crates, and only leaving them at `pubnub-hyper`.

To deliver this, it introduces two new traits: `Transport` and `Runtime`, each encapsulating the corresponding responsibilities. `pubnub-core` provides a generic implementation of the client that relies of those traits, `pubnub-hyper` 

It also refactors the subscription-related parts of the crate and adds enhances integration tests with test cases to catch more issues with cloned clients. Overall operation logic generally remains the same, except few internal changes that are well documented.


pubnub-core/src/subscription/registry.rs
```diff
@@ -0,0 +1,72 @@
...
+            self.map.remove(name);
+            UnregistrationEffect::NameErazed
+        } else {
+            UnregistrationEffect::NamePreserved
+        };
+
+        Some((removed, effect))
+    }
+
+    // TODO: provide better interface for iteration over mutable items.
```
parasyte: Any specific ideas in mind?

pubnub-core/src/subscription/subscribe_loop.rs
```diff
@@ -0,0 +1,282 @@
...
+use log::{debug, error};
+
+pub(crate) use super::channel::{Rx as ChannelRx, Tx as ChannelTx};
+pub(crate) use super::registry::ID as SubscriptionID;
+
+pub(crate) type Registry = GenericRegistry<ChannelTx>;
+
+pub(crate) type ReadyTx = oneshot::Sender<()>;
+#[allow(dead_code)]
+pub(crate) type ReadyRx = oneshot::Receiver<()>;
```
parasyte: I would prefer not allowing dead code. Let's not maintain this. YAGNI.

pubnub-core/src/subscription/subscribe_loop.rs
```diff
@@ -0,0 +1,282 @@
...
+    } = params;
+
+    let mut encoded_channels = EncodedChannelsList::from(&channels);
+    let mut encoded_groups = EncodedChannelsList::from(&groups);
+
+    let mut timetoken = Timetoken::default();
+
+    loop {
+        // TODO: do not clone cached values cause it defeats the purpose of the
+        // cache. Do we need this cache though? There are no benchmarks...
```
parasyte: This clone can be removed if `subscribe_request::subscribe_request` doesn't borrow the string for the lifetime of its return value.

As for whether the cache is necessary, my gut instinct tells me that any serde (percent-encoding, JSON, etc) is a huge time sink. And I'm happy to spend a little memory to free the CPU for more important tasks.

pubnub-core/src/subscription/subscribe_loop.rs
```diff
@@ -0,0 +1,282 @@
...
+/// Encodes action to be taken in response to control command.
+#[derive(Debug)]
+enum ControlOutcome {
+    Terminate,
+    CanContinue,
+}
+
+/// Handle a control command.
+///
+/// This is split out from the `select!` macro used in the subscribe loop.
```
parasyte: The caller is no longer using `select!` :+1:

pubnub-core/src/subscription/subscribe_loop_supervisor.rs
```diff
@@ -0,0 +1,161 @@
...
+                let subscribe_loop_params = SubscribeLoopParams {
+                    control_rx,
+                    ready_tx: Some(ready_tx),
+                    exit_tx: self.params.exit_tx.clone(),
+
+                    transport: pubnub.transport.clone(),
+
+                    origin: pubnub.origin.clone(),
+                    agent: pubnub.agent.clone(),
+                    subscribe_key: pubnub.subscribe_key.clone(),
```
parasyte: Shouldn't these just be cloned into the `SubscribeLoopSupervisor` struct along with `pubnub.runtime`? I'm worried that the new interface, accepting a reference to a `PubNub` struct, could be unsound if a different struct reference is passed in.

pubnub-core/src/subscription/subscribe_loop_supervisor.rs
```diff
@@ -0,0 +1,161 @@
...
+            let id_or_retry = if let Some(ref mut control_tx) = self.control_tx {
+                // Send a command to add the channel to the running
+                // subscribe loop.
+
+                debug!("Adding channel {:?} to the running loop", channel);
+
+                let (id_tx, id_rx) = oneshot::channel();
+
+                // TODO: unify interfaces to either use `ListenerType` or
+                // `&str` when we refer to a channel.
```
parasyte: Yes. This method needs to be able to create a `Subscription` for either a channel or a channel-group.

pubnub-core/src/subscription/subscription.rs
```diff
@@ -0,0 +1,95 @@
...
+        Stream::size_hint(&self.channel_rx)
+    }
+}
+
+impl<TRuntime: Runtime> Subscription<TRuntime> {
+    /// Request to unsubscribe the subscription from the channel.
+    ///
+    /// This call completes when the unsubscription request is submitted
+    /// successfully, but doesn't indicate when the unsubscription request
+    /// is processed by the subscribe loop control command handler.
```
parasyte: Why is this necessary? `std::mem::drop(subscription);` does the same thing.

pubnub-core/src/subscription/mod.rs
```diff
@@ -0,0 +1,15 @@
+mod encoded_channels_list;
+mod mvec;
+mod registry;
+mod subscribe_request;
```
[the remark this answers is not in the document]
MOZGIII: There's a lot of reorganization pending, I'll be moving stuff around more later.
Just to double-check I understand this correctly, your suggestion is to put all the mods above into the `subscribption.rs` file, right?

pubnub-core/src/subscription/registry.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Debug)]
+pub enum RegistrationEffect {
+    NewName,
+    ExistingName,
+}
+
+#[derive(Debug)]
+pub enum UnregistrationEffect {
+    NameErazed,
```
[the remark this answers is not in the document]
MOZGIII: Right, my bad.

pubnub-core/src/subscription/registry.rs
MOZGIII: There are other todos at use locations that provide more hints. In essence, I'd replace this with just an iterator over the `MVec` values.

pubnub-core/src/subscription/subscribe_loop.rs
MOZGIII: I agree. The reason I left those like that is cause I was considering slicing the crate into the smaller chunks (making those parts of the public API), but ended up not doing that.

pubnub-core/src/subscription/subscribe_loop.rs
MOZGIII: > This clone can be removed if `subscribe_request::subscribe_request` doesn't borrow the string for the lifetime of its return value.

Right! I'll separate the lifetimes, this should solve this issue.

> As for whether the cache is necessary, my gut instinct tells me that any serde (percent-encoding, JSON, etc) is a huge time sink. And I'm happy to spend a little memory to free the CPU for more important tasks.

I absolutely agree with you about the gut instinct, I have this too. However, it looks like yet another premature optimization case. I don't mind it now, but I think this part can safely be removed if it causes flexibility issues when we switch from using `url`s in `Transport`.

pubnub-core/src/subscription/subscribe_loop.rs
MOZGIII: Yep, I have to clean this up.

pubnub-core/src/subscription/subscribe_loop_supervisor.rs
MOZGIII: As I see, the channel/channel-group abstraction was not implemented to completion when I started, otherwise, it'd be integrated globally and I wouldn't be problematic here.
I don't want to solve this within this PR, but it can definitely get some polishing later on.

pubnub-core/src/subscription/subscribe_loop_supervisor.rs
MOZGIII: Yes, this is one of the issues that unraveled toward the very end of the time frame that I had to work on this, so it's a bit incomplete. In the future, I'd take some time to eliminate extra clones and re-think the ownership model of the entities in a broader sense. I.e., the `origin`, `agent` and `subscribe_key` should most likely belong to the `Transport` implementation itself (maybe not directly, via some common struct for HTTP-based transports) - that way we don't have to even deal with them at this API layer. This would be a part of getting rid of the `url`s in `Transport` API. In fact, I sort of have a plan on how to continue from here on to further improvements. Eliminating references to `PubNub` is one of the major concerns of mine, but there are some blockers that need to be unlocked first before this can be done cleanly.

pubnub-core/src/subscription/subscription.rs
MOZGIII: With `drop` you can't wait for the actual drop to happen. From the use case perspective, this might be necessary to ensure no network requests are issued by the subscribe loop after a certain point. Using `drop ` would not be sufficient since it does not wait for the command to complete.

Having both `Drop` impl and this fn covers both resource leakage prevention when user doesn't care about the precise order of the events (i.e. doesn't need to ensure the subscribe loop processes drop before the user code continues), and the case when it matters for the user.

A few days ago, I was thinking about this implementation and I've spotted an issue. The `Subscription` struct should track and only send a single unsubscription request. I deem this should be implemented by a better abstraction - like a consumable one-shot unsubscription handle for `Subscription` to own, atop of the mpsc control queue.

pubnub-core/src/pubnub.rs
```diff
@@ -174,123 +183,104 @@ impl PubNub {
...
+{
+    /// Build the PubNub client to begin streaming messages.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use pubnub_hyper::PubNubBuilder;
+    ///
+    /// let pubnub = PubNubBuilder::new("demo", "demo")
+    ///     .build();
```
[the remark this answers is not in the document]
MOZGIII: Right, simply properly documenting these would eliminate the odd dependency on the `pubnub-hyper`. However, the price for that, in the current situation, would be making the example `ignore`, and write some non-buildable code that users won't be able to copy-paste and use.
On the other hand, this is the most accurate example code we can provide to the users at the moment. This is why I decided to go with this for now.
A lot of the values in this builder are not used anyhow, so we might be better rethinking the core approach and clearing things up.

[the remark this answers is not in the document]
parasyte: It looks like the `pubnub-core` crate is not meant to be used directly by users, anyway. So documentation would target e.g. implementers of higher level crates like what `pubnub-hyper` is to Tokio.

pubnub-core/src/subscription/mod.rs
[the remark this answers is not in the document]
parasyte: Actually this comment was meant to address a misconception I had while reading through the code. For some reason I saw these as `use` statements. Nothing to see here. ;)

pubnub-core/src/subscription/registry.rs
parasyte: Not entirely sure I follow. The only use of the iterator is accessed through a `HashMap<MVec<T>>`, so the call site needs to gain access to it through the `HashMap` interface. Is the idea here to combine the two interfaces?

pubnub-core/src/subscription/subscribe_loop.rs
parasyte: I really like where this is going. It will DRY up a great deal of the data interfaces in here.

pubnub-core/src/subscription/subscribe_loop_supervisor.rs
parasyte: I really like where this is going. It will DRY up a great deal of the data interfaces in here.

pubnub-core/src/subscription/subscribe_loop.rs
parasyte: Just another gut feeling here, but it seems like the cache would just have to move closer to the `Transport` if the types were improved. Not necessarily that removing it entirely would be ideal.

pubnub-core/src/subscription/subscribe_loop_supervisor.rs
parasyte: The incompleteness was partially my fault. Not making excuses, but I just didn't need the channel groups abstraction at the time.

Fair, we don't have to make any unrelated improvements in this PR. I was only commenting on the new comment added in the PR.

pubnub-core/src/subscription/subscription.rs
parasyte: There might be a use case for waiting for the network layer to do its thing, but I haven't ever found one in practice. Perhaps this is more important in resource-constrained environments?

Once dropped, messages can no longer be received on that channel, even if the network layer (and subscription manager) manage to receive anything after the drop. This case is gracefully handled, and the end user doesn't need to be aware of it.

A big part of the challenge is actually in adapting a streaming interface over a multiplexed HTTP-based protocol. After working with this specific protocol for nearly a decade, I'm pretty confident that we really _don't_ want to allow unfortunate parts of this protocol abstraction to leak into the streaming interface.

The other point I want to make is that Rust doesn't have an asynchronous `Drop`, and it might never happen. But the need for it has at least been identified: https://boats.gitlab.io/blog/post/poll-drop/ see also discussion on async cancelation, which is very similar in concept and necessity: https://internals.rust-lang.org/t/async-await-the-challenges-besides-syntax-cancellation/10287 (the linked gist has several relevant details).

Assuming something like either of these could be relied upon, I would lean heavily toward just dropping the stream when a subscription needs to be, quite literally, dropped. At least for now, since it's the model that most closely fits with existing stream mechanisms in the Rust ecosystem. This might change over time as the ecosystem evolves.

MOZGIII: I'm actually using in the tests already to orchestrate the behavior of the control loop into edge cases. This kind of workaround is pretty common, it is indeed specific to the async drop absence.

There are two ways we could implement drop: with blocking internal cleanup operation, and with non-blocking, but offloaded to a separately scheduled `Task`. The first approach makes it impossible to use the implementation in the async context because blocking would be critically problematic for the whole runtime. So, the second way is typically used, but it has another problem - you can't wait for the actual resource cleanup completion. For this reason, async `Drop` implementers typically also provide the async function that returns a future which resolves after the resource allocation if fully done (or similar reasonable logical even is reached).

Practically, the cancellation is already effectively solved by the way we handle commands in the control loop, so all we need is to just send the drop command. From the interface point of view, the `Subscription` struct is our cancellation token, so we can drop it (or issue an unsubscription in an async-firedly way as implemented currently).

Actually, it may be problematic to have an async operations in the drop in the first place. A trivial (and unoptiomal - more on this later) way of walking around it would be to pass the drop requests not via the typical control tx we usually use, but via an `oneshot` channel, sending over which is not async. This would require maintaining a `oneshot` channel per subscription, which has a lot of overhead. If we go with a lower-level implementation, we can make a specialized custom channel to do this optimally: send without blocking, and receive from multiple producers with just a single waker. I have experience with implementing these sorts of things, and will be able to implement one, however it might not be high in the priority list.

To sum up, there are multiple ways we can change the current implementation, and, I added the `unsubscribe` call to actually solve a pain point that wasn't covered otherwise - precise control over the control loop in the tests. We can approach the solution for that pain point differently, and eliminate the very need for this `unsubscribe` call - for example via additional control loop hooks. I expect more options to unravel as we continue refactoring other parts further. For this reason, I'd propose to just postpone the decision on this API call for now - if we remove it now, it'd be troublesome to support some of the test cases (among the most important ones, since the ones in question did actually caught bugs), and if we implement something else and the refactor other parts we might end up changing this again after that cause there will be a better option. So I'd recommend to just keep in mind for now that we might want to get rid of it.

parasyte: > So, the second way is typically used, but it has another problem - you can't wait for the actual resource cleanup completion.

I'm honestly happy with this tradeoff for now. Like I alluded earlier, there is probably a use case for waiting, I just haven't seen it myself in practice. As long as we can guarantee that resource cleanup will be performed when necessary, the cleanliness of the `Drop` approach feels much more natural.

I'm aware that `Drop` cannot always be guaranteed, because of the reference cycle problem (and panic unwinds and other related issues). My argument here is that if this behavior is acceptable to Rust, it should be acceptable to any crate written in Rust. Memory leaks are not memory safety violations.

> Practically, the cancellation is already effectively solved by the way we handle commands in the control loop, so all we need is to just send the drop command. From the interface point of view, the Subscription struct is our cancellation token, so we can drop it ...

Exactly. 

> (or issue an unsubscription in an async-firedly way as implemented currently).

With the challenge that it's a bit of an anti-pattern that will allow a dead `Subscription` to be used in some other logical thread, including whatever error handling may be necessary for handling this new edge case.

A synchronous drop would be awesome if it can be done (it's what I wanted all along). Feels like it will almost certainly have to cross a thread boundary, though. I'm wary of that issue. And that's why `Drop` reuses the `channel` at present.

MOZGIII: We can definitely improve a way the `unsubscription` is implemented with the current approach, as has a few bugs, as I said above.
If you’re worried about this being an anti-pattern though, I would disagree. And it is not about memory safety or leaks, it is about eliminating race conditions. The similar functionality is commonly provided across the rust ecosystem:
- https://docs.rs/futures/0.3.1/futures/channel/mpsc/struct.Receiver.html#method.close
- https://docs.rs/tokio/0.2.8/tokio/sync/mpsc/struct.Receiver.html#method.close
- similar stuff for sockets and so on

> I'm honestly happy with this tradeoff for now. Like I alluded earlier, there is probably a use case for waiting, I just haven't seen it myself in practice. As long as we can guarantee that resource cleanup will be performed when necessary, the cleanliness of the `Drop` approach feels much more natural.

I added the `unsubscribe` strictly for practical reasons, there are a couple new tests that simply can’t be implemented with just the `Drop`, exactly because it causes a race condition (and in fact don’t produce correct results if `drop(s)` is used instead of `s.unsubscribe().await`.

It is true though that we could make `unsubscribe` take `self` by value - to simplify the interface. Currently, it is still possible to read messages after the unsubscription request was submitted - because there’s a race between when us sending a command to control loop and the http request completion. Technically, we could end up with new messages being available in the subscription, while the subscription has the drop command pending on the control loop. User might want to handle this case properly - and it’s reasonable to let this, since the compute resources have been spent, and we have the meaningful results. The beauty here is it’s the user’s discretion whether he wants to handle that case or not. If one doesn’t, a usual drop will suffice, and the there’s practically no need to use this method for those use cases.

Take a look at this commit: https://github.com/pubnub/rust/commit/254df9672862fdee3742812ab923508f2dfa7e83
There, the main value is the test itself, as it does check some interesting stuff. However, to be able to actually implement it I *had to* add the `unsubscribe` call. So, if we were to change this, we’d have to deal with the test breakage too. The test is important though. It is nothing we can’t solve otherwise somehow, but addressing it a few steps later might be easier, cause we’ll be able to focus solely on this issue with other things out of the door.

parasyte: I must admit, I don't understand the purpose of the linked test. The transport layer should be safe to cancel at any time. I get that there are race conditions, but most race conditions in this code are the "don't care" variety.

Could you provide more details about why it matters that the subscribe loop can be terminated while an HTTP request is outstanding? Or if I've misunderstood the intention, could you clarify the intention of the test?

MOZGIII: The meaningful part is here: https://github.com/pubnub/rust/commit/254df9672862fdee3742812ab923508f2dfa7e83#diff-6ab62c8d2d967ad2931b0a62a8186475R166-R194

This integration test ensures there're no issues with the subscribe loop termination and restart logic. It doesn't really test the transport (at least that wasn't the intention).

I'm reading that test code now, and apparently, eliminating the `unsubscribe` and just dropping the subscription shouldn't cause any issues. I apologize, I think I've got confused somehow - I was under the impression it was required.

Now, I don't think there's a real internal use case for the `unsubscribe`. Given that it's kind of redundant, I don't mind removing it. We can add it again later if there's a need.

parasyte: Nice! Yeah, I think we are in agreement on this one now.

pubnub-core/src/subscription/registry.rs
MOZGIII: Pretty much, yes. We don't really need to expose the internal `MVec<T>` in the interface, all we need is to be able to enumerate the listeners for a certain route:

https://github.com/pubnub/rust/blob/dd3729ef2b9737b6c7ba008bd31415900e743ef1/pubnub-core/src/subscription/subscribe_loop.rs#L180-L182

This is the only call-site we currently have.

pubnub-core/src/subscription/subscribe_loop.rs
MOZGIII: Oh, yeah, I think that'd be the case! Good point!

pubnub-core/src/pubnub.rs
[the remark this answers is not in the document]
MOZGIII: This is temporary until we move the unrelated variables to transports. After that, we won't have to keep the unrelated values in the builder or in the "core" `PubNub` component altogether. We might even add a more advanced builder to the "hyper" to match the current interface.

pubnub-core/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+//! # Async PubNub Client SDK for Rust
+//!
+//! - Fully `async`/`await` ready.
+//! - Uses Tokio and Hyper to provide an ultra-fast, incredibly reliable message transport over the
```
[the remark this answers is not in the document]
MOZGIII: I'm working on a core crate doc rewrite - it'll be more accurate in terms of what it offers after the refactoring.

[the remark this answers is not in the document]
MOZGIII: This should be better now. I used some of [this goodness](https://github.com/rust-lang/rust/issues/43466), but I'm not sure if it's rolled out to the docs.rs already. I guess we'll have to push a version to see if it works.

pubnub-core/src/pubnub.rs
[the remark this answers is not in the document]
MOZGIII: I propose we leave it as is for now. The builder is reexported, so it might still be useful for the users until we polish the builder. What do you think, @parasyte?

pubnub-core/src/subscription/encoded_channels_list.rs
```diff
@@ -0,0 +1,37 @@
+use super::registry::Registry;
+use percent_encoding::{utf8_percent_encode, NON_ALPHANUMERIC};
+
+// TODO: eliminate this after we switch Transport to passing structured values
+// instead of the URLs.
```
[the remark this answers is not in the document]
MOZGIII: I was planning on doing this outside of this PR, to limit the number of changes introduced at once. There's a lot of TODOs, but going over all of them would better be done in separate PRs.

pubnub-core/src/subscription/subscribe_loop.rs
MOZGIII: I updated the TODO to capture the decision to move the cache to the transport.

MOZGIII:
Almost everything brought up so far in the review is addressed this way or another! :tada: 

Do you think we can move forward with this? I'd appreciate more review passes.

pubnub-core/src/pubnub.rs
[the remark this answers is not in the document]
parasyte: I'd like to go over the code again before I give a definitive answer, but I'm leaning toward "yes" for now. If nothing else, maybe the `use` lines could be updated?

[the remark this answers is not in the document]
MOZGIII: I'm not sure I understand what you mean. They're not invalid - doctests are passing...
The alternative we have there is to simply remove the examples. We can also give another example and give mark the code section `ignore`, but I'd prefer if all the examples are tested - that way they won't get stale over time.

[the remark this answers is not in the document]
parasyte: I know the tests are valid. It just looks weird to be importing `pubnub_hyper` in the docs for `pubnub_core`. If changing the crate name in the `use` line is unacceptable, then move the doctests to the `pubnub_hyper` crate?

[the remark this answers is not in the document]
MOZGIII: We could do something like this:

```rust
    /// # use pubnub_hyper::PubNubBuilder;
    /// let pubnub = PubNubBuilder::new("demo", "demo")
    ///     .build();
```

Then it won't involve the `pubnub_hyper` crate. There's a downside though - users would most likely be confused trying to reproduce that example.

I can try some other ways to eliminate the `pubnub_hyper` crate from the doctests. My thinking was, even with the explicit `pubnub_hyper` in, it's the best example we can give to the users in the current context (no magic, copy-pastable stuff), yet it's true it's far from ideal if we have to reference another crate. So, yeah, I'll just try a few other things. :smile:

parasyte:
Ok, I'm getting back to looking at this again. I will have time for a thorough review, so we can expect some feedback on all the things. :)

pubnub-core/src/subscription/subscription.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+fn is_drop_send_result_error(result: Result<(), mpsc::SendError>) -> bool {
+    match result {
+        Ok(_) => false, // not an error
+        Err(err @ mpsc::SendError { .. }) if err.is_disconnected() => {
+            // Control handle is dead, so we can assume loop is dead
+            // too, and that we have, therefore, kind of unsubscribed
+            // successfully.
```
parasyte: :+1: Yep, seems about right!

pubnub-core/src/tests/mock/transport.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use mockall::mock;
+
+#[derive(Debug, Error)]
+#[error("mock tranport error")]
+pub struct MockTransportError;
+
+mock! {
+    pub Transport {
+        #[allow(clippy::type_complexity)]
```
parasyte: nitpick: :sob:

Is this really necessary, or was it leftover from development? Some extra type definitions would really clean up this code.

pubnub-core/src/transport.rs
```diff
@@ -0,0 +1,18 @@
...
+#[async_trait]
+pub trait Transport: Clone + Send + Sync {
+    /// Transport-specific error type this transport can generate.
+    type Error: std::error::Error + Send + Sync;
+
+    /// Send a publish request and return the JSON response.
+    async fn publish_request(&self, url: Uri) -> Result<Timetoken, Self::Error>;
+
+    /// Send a subscribe request and return the JSON messages received.
+    async fn subscribe_request(&self, url: Uri) -> Result<(Vec<Message>, Timetoken), Self::Error>;
```
parasyte: Not expecting any outcome on this comment, just want to leave feedback that it would be super cool if this trait wasn't hardcoded to `http`. We're investing in some additional protocols; MQTT is already available, and gRPC is on the way. Having a gRPC transport would be incredible!

pubnub-hyper/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+/// A sensible default variant of the hyper runtime.
+pub use crate::transport::hyper::Hyper as DefaultTransport;
+
+use crate::core::{PubNub as Core, PubNubBuilder as CoreBuilder};
+
+/// PubNub client bound with hyper transport and tokio runtime.
+pub type PubNub = Core<DefaultTransport, DefaultRuntime>;
+
+/// PubNubBuilder bound with hyper transport and tokio runtime.
+pub type PubNubBuilder = CoreBuilder<DefaultTransport, DefaultRuntime>;
```
parasyte: This is where the documentation challenge is biting me. I don't know if rustdoc has a good way of importing docs from another crate (`pubnub-core` in this case), or if there are better options...

But at a minimum, we need a dumb example with `PubNub::new()` and `PubNubBuilder::new()` here, and links to `pubnub_core::PubNub` and `pubnub_core::PubNubBuilder` for the method docs.

Our SDK documentation has historically been very bad, and I want to do everything we can to improve the situation, this time. I honestly don't know if our docs team will care about the rustdoc generation, but anyone who finds it on `docs.rs` will appreciate good documentation.

pubnub-core/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+//! # PubNub Core
+//!
+//! Provides the common high-level logic for PubNub clients.
+//!
+//! - Fully `async`/`await` ready.
+//! - Modular, bring your own [`Transport`] and [`Runtime`].
+//! - Multiplexes subscription polling for multiple logical streams over a
+//!   single transport invocation to optimize kernel network subsystem pressure.
+//!
+//! Build your own client, or use preconfigured [`pubnub-hyper`](pubnub_hyper).
```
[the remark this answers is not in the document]
MOZGIII: This will be resolved when we release the creates. I'll take care of it, it might require a set of commits to get around.

pubnub-core/src/tests/mock/transport.rs
MOZGIII: You're right! I'll either take care of that at some point or add a comment explicitly justifying the need for that exception.

PR #34, opened 2020-02-25
Presence

MOZGIII:


pubnub-core/src/data/presence.rs
```diff
@@ -0,0 +1,82 @@
...
+pub mod respond_with {
+    //! Type system level flags to specialize the response types.
+
+    /// Return occupancy only.
+    #[derive(Debug, Clone, Copy)]
+    pub struct OccupancyOnly;
+    /// Return occupancy and UUIDs of the users.
+    #[derive(Debug, Clone, Copy)]
+    pub struct OccupancyAndUUIDs;
+    /// Return cooupance, UUIDs of the users and the related states.
```
parasyte: nitpick, typo:

```suggestion
    /// Return occupancy, UUIDs of the users and the related states.
```

pubnub-core/src/data/presence.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    impl RespondWith for OccupancyOnly {
+        type Response = ChannelInfo;
+    }
+    impl RespondWith for OccupancyAndUUIDs {
+        type Response = ChannelInfoWithOccupants<UUID>;
+    }
+    impl RespondWith for Full {
+        type Response = ChannelInfoWithOccupants<ChannelOccupantFullDetails>;
+    }
```
parasyte: Wondering your thoughts, here... Could this use a struct-like enum instead of traits?

pubnub-hyper/src/transport/hyper/presence.rs
```diff
@@ -84,9 +180,252 @@ impl TransportService<request::GetState> for Hyper {
...
+        request: request::HereNow<presence::respond_with::OccupancyOnly>,
+    ) -> Result<Self::Response, Self::Error> {
+        let request::HereNow {
+            channels,
+            channel_groups,
+            ..
+        } = request;
+
+        let channels = EncodedChannelsList::from(channels);
+        let channel_groups = EncodedChannelsList::from(channel_groups);
```
parasyte: Something to watch out for: DRY.

pubnub-core/src/data/presence.rs
MOZGIII: Thanks! I noticed this before, but couldn't find it. :+1:

pubnub-core/src/data/presence.rs
MOZGIII: The core idea here is to have this handled at the type system level. District types allow us to do this, while having an enum, on the contrary, doesn't allow us to involve the type system. We rely on this to shape the inputs and outputs of the `Transport` service.

PR #43, opened 2020-03-22
Eliminate duplicate linting instructions and reorganize for clarity

MOZGIII:


MOZGIII:
This must be a Travis' internal problem. Please merge this one for me! :smile:

parasyte:
The Travis CI problem is blocking merge. Will have to get this sorted out of band.

MOZGIII:
I can repush this to retrigger the build, maybe that'll help

MOZGIII:
It might be that the branch test is simply not invoked for non-owners. It might be a security policy.

cc @davidnub

davidnub:
Actually it looks related to `clippy` on AMD64

image

MOZGIII:
That's from the successfully completed job, the problem is the one that's stuck:
image

davidnub:
> That's from the successfully completed job, the problem is the one that's stuck:
> image

Yea sorry, you're right. I think perhaps it's related to Travis trying to run on your branch of your repo. I wonder if you set that up to run with the free Travis account if that would solve it?

davidnub:
Or if you add @parasyte or I as maintainers to your repo fork then we can add it to our Travis run.

MOZGIII:
I got my access restored, so I can work around this by recreating the PR from a branch.

> Or if you add @parasyte or I as maintainers to your repo fork then we can add it to our Travis run.

This is a rather odd requirement for a PR. I understand where it comes from, but it seems like it'll create a serious boundary for the open-source community contributors.

In the hindsight, we probably should reconfigure the merge requirement to omit the check for branch tests to pass.

davidnub:
Great that you got your access restored.

> This is a rather odd requirement for a PR. I understand where it comes from, but it seems like it'll create a serious boundary for the open-source community contributors.

I agree that this is odd and shouldn't be necessary. We have a different process in the works when it comes to the internal and external development workflow which will hopefully address some of these challenges going forward.

Issue #52, opened 2020-04-23
Isolate integration tests


MOZGIII:
Currently, running integration tests can fail due to the improper state of the PubNub network.

See [this](https://travis-ci.com/github/pubnub/rust/jobs/322519085) CI failure for example.

It was kind of ok before, but now we're starting to add more and more features, some of which require exclusive access to global state to assert properly.

We should figure out a way to isolate integration tests.

One other solution is to ditch integration tests against real infrastructure, but testing against the real infrastructure definitely has it's pros compared to testing against thew mocked infrastructure, and the technical difficulties shouldn't justify ditching it.

Related: #36.

MOZGIII:
A follow up from our sync up call on this matter: we decided that using a like VCR to record server responses and emulate them would be the best way do go, for now. Other solutions aren't as good or even possible.

Issue #36, opened 2020-03-04
Switch pubnub-hyper integration tests to httptest


MOZGIII:
Currently, `pubnub-hyper` is using real PubNub network to perform conduct tests. This may be unnecessary, and also limiting, cause we can't test failure scenarios.

Switching to [httptest](https://github.com/ggriffiniii/httptest) should solve that issue.

PR #58, opened 2020-05-03
Re-enable CI failures

MOZGIII:
This should be merged once we fix the CI and tests.

Blocked by #52.


MOZGIII:
As discussed, I'm merging this.

We still haven't resolved the test reliability issues (see #52 and #36). However, since the development is paused, it's better to have tests work properly (i.e. fail on errors), even if the tests are sometimes failing. This way, when the development is resumed, there won't be an unpleasant surprise of tests failing silently. At least they'll fail properly.

PR #59, opened 2020-05-31
History API

MOZGIII:
Implements history API.

Tests are not passing due to delete request being unavailable for demo account.

### DIFF
--- a/pubnub-core/src/data/request.rs
+++ b/pubnub-core/src/data/request.rs
@@ -8,7 +8,7 @@ use crate::data::presence;
 use crate::data::pubsub;
 use crate::data::timetoken::Timetoken;
 use crate::data::uuid::UUID;
-use std::marker::PhantomData;
+use std::{collections::HashMap, marker::PhantomData};
 
 /// A request to publish a message to a channel.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -202,5 +202,5 @@ pub struct MessageCountsWithTimetoken {
 pub struct MessageCountsWithChannelTimetokens {
     /// A list of channels with timetokens to get message counts at.
     /// Timetoken value must be non-zero.
-    pub channels: Vec<(channel::Name, history::Timetoken)>,
+    pub channels: HashMap<channel::Name, history::Timetoken>,
 }
--- a/pubnub-core/src/data/request.rs
+++ b/pubnub-core/src/data/request.rs
@@ -1,5 +1,6 @@
 //! Types used by [`crate::Transport`].
 
+use super::history;
 use crate::data::channel;
 use crate::data::object::Object;
 use crate::data::pam;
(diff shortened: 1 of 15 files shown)


parasyte:
We might be able to enable history delete on `demo`. If not, I can create a key-pair specifically for this test!

pubnub-hyper/src/transport/hyper/history.rs
```diff
@@ -0,0 +1,245 @@
...
+        let request::MessageCountsWithChannelTimetokens { channels } = request;
+
+        // Tranform input to the proper format.
+        let cap = channels.len();
+        let mut names = Vec::with_capacity(cap);
+        let mut timetokens = Vec::with_capacity(cap);
+        for (channel, timetoken) in channels.into_iter() {
+            names.push(channel);
+            timetokens.push(timetoken);
+        }
```
parasyte, line 157: Ah! This can be made simpler with a little Functional love.

```suggestion
        // Unzip Vector-of-Tuples into separate Vectors.
        let (names, timetokens): (Vec<_>, Vec<_>) = channels.into_iter().unzip();
```

Also works with `HashMap`

pubnub-core/src/data/request.rs
```diff
@@ -122,3 +123,84 @@ pub struct Heartbeat {
...
+    /// Must be greater than zero.
+    pub timetoken: history::Timetoken,
+}
+
+/// Get message counts over a time period per channel.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MessageCountsWithChannelTimetokens {
+    /// A list of channels with timetokens to get message counts at.
+    /// Timetoken value must be non-zero.
+    pub channels: Vec<(channel::Name, history::Timetoken)>,
```
parasyte, line 205: Nitpick: This could also be a `HashMap<channel::Name, history::Timetoken>`

pubnub-util/src/uritemplate.rs
```diff
@@ -48,21 +48,21 @@ impl UriTemplate {
...
-    pub fn set_list<T: Into<String>>(
+    pub fn set_list<T: ToString>(
         &mut self,
         varname: impl AsRef<str>,
         var: impl IntoIterator<Item = T>,
     ) -> &mut Self {
         self.0.set(
             varname.as_ref(),
-            TemplateVar::List(var.into_iter().map(Into::into).collect()),
+            TemplateVar::List(var.into_iter().map(|val| val.to_string()).collect()),
```
parasyte, line 58: `ToString::to_string` ought to work here.

```suggestion
            TemplateVar::List(var.into_iter().map(ToString::to_string).collect()),
```

parasyte:
On the failing test, can you try signing the request? Now that PAM is supported, this should just be a matter of including the secret key from the env ... I think that will unblock us here.

pubnub-util/src/uritemplate.rs
MOZGIII: There's a lifetime mismatch there, so it doesn't work. I tried doing it like that too at first.

MOZGIII:
How would the signing work in this case? We generate the request, then sign it, and then send it along with the signature at the `auth` field?
We might have to rework how the transport is structured to allow signing calls like this - to extract the request preparation and parsing parts out so we can either issue the HTTP request directly, or sign it and the issue it. Such changes should've been a part of the PAM feature, and this means we're really not done with that yet.

Can this test be implemented without signing though? If it can't be, we might need to hold on with this PR before we do the PAM thing.

pubnub-util/src/uritemplate.rs
parasyte: Oof! That's too bad. No big deal.

parasyte:
Signing is the same process supported by the grant API. In our discussion this morning, we talked about what it would take to implement request signing on all requests in the transport. A backlog ticket should be created for this.

I have enabled the feature on the demo key, and restarted the CI job. The server now delivers an HTTP 200 response, but the test assertion fails. I think that's the last hangup with this.

MOZGIII:
Created #60.

PR #63, opened 2022-02-14
test(github-actions): migrate tests to GitHub Actions

parfeon:
test(github-actions): migrate tests to GitHub Actions

Migrate PubNub SDK test suite from Travis to GitHub Actions.


stmpjmpr:
@parfeon I noticed that some of the failures here are fixed in #65. Try merging that and re-running these tests.

PR #65, opened 2022-02-25
Fix travis build

stmpjmpr:
I made changes to fix the linter/Clippy errors, as well as formatter errors that were causing the build to fail. Important to note:

- Minimum Rust version is now 1.49.0, due to features that cause warnings in 1.46.0 now.
- I widened what is allowed for unknown, renamed, and removed lints, due to a changed linter name that either breaks the minimum version, or the current version, mutually exclusive. I added TODOs noting when we can re-enable the check in the future.
- There were quite a few places where unnecessary dereferences were taking place, and semicolons added where no value was being returned and used.

### DIFF
--- a/pubnub-core/src/lib.rs
+++ b/pubnub-core/src/lib.rs
@@ -10,6 +10,7 @@
 //! Build your own client, or use preconfigured [`pubnub-hyper`](pubnub-hyper).
 
 #![allow(unknown_lints)]
+#![allow(renamed_and_removed_lints)]
 #![deny(
     clippy::all,
     clippy::pedantic,
--- a/pubnub-core/src/lib.rs
+++ b/pubnub-core/src/lib.rs
@@ -17,6 +17,7 @@
     missing_copy_implementations,
     rustdoc::broken_intra_doc_links
 )]
+#![allow(unknown_lints)]
 #![allow(clippy::doc_markdown)]
 #![forbid(unsafe_code)]
 
--- a/pubnub-core/src/lib.rs
+++ b/pubnub-core/src/lib.rs
@@ -9,8 +9,11 @@
 //!
 //! Build your own client, or use preconfigured [`pubnub-hyper`](pubnub-hyper).
 
+// TODO: Remove these when minimum Rust version >1.59.0, when the name changed.
+// TODO: `broken_intra_doc_links` below should become `rustdoc::broken_intra_doc_links`.
 #![allow(unknown_lints)]
 #![allow(renamed_and_removed_lints)]
+
 #![deny(
     clippy::all,
     clippy::pedantic,
--- a/pubnub-core/src/lib.rs
+++ b/pubnub-core/src/lib.rs
@@ -15,7 +15,7 @@
     missing_docs,
     missing_debug_implementations,
     missing_copy_implementations,
-    broken_intra_doc_links
+    rustdoc::broken_intra_doc_links
 )]
 #![allow(clippy::doc_markdown)]
 #![forbid(unsafe_code)]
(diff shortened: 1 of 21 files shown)


pubnub-core/src/lib.rs
```diff
@@ -9,6 +9,10 @@
 //!
 //! Build your own client, or use preconfigured [`pubnub-hyper`](pubnub-hyper).
```
phairow, line 11: I'm not sure I fully understand the comments here. What changed in 1.59? Is it just the location of the lint rule that changed?

pubnub-core/src/mock/runtime.rs
```diff
@@ -29,6 +29,6 @@ impl Runtime for MockRuntime {
     where
         F: Future<Output = ()> + Send + 'static,
     {
-        self.mock_workaround_spawn(Box::pin(future))
+        self.mock_workaround_spawn(Box::pin(future));
```
phairow, line 32: I'm still getting used to the way semicolon works in rust. From my understanding it suppresses the result of the expression from getting returned. What is happening here?

pubnub-core/src/subscription/subscribe_loop.rs
```diff
@@ -161,6 +161,7 @@ enum ControlOutcome {
 }
 
 /// Handle a control command.
```
phairow, line 163: I wonder if we can get rid of this exception in the future by adding an async test or example.

pubnub-core/src/subscription/subscription.rs
```diff
@@ -55,9 +59,10 @@ impl<TRuntime: Runtime> Drop for Subscription<TRuntime> {
         // See: https://boats.gitlab.io/blog/post/poll-drop/
         self.runtime.spawn(async move {
             let drop_send_result = control_tx.send(command).await;
-            if is_drop_send_result_error(drop_send_result) {
-                panic!("Unable to unsubscribe");
-            }
+            assert!(
+                !is_drop_send_result_error(drop_send_result),
```
phairow, line 63: I don't know why this is confusing me so badly but I think I understand now. Looks like assert will panic when the expression is true so this should be equivalent to the if statement. I'm still learning :)

phairow (approved):
LGTM! I had some questions since I am new to rust but it looks good overall.

pubnub-core/src/lib.rs
stmpjmpr: The "broken_intra_doc_links" rule's name changed to "rustdoc::broken_intra_doc_links. The older build doesn't recognize the new name, and vice versa. We have the linter rules strict enough that this, which would normally be a warning, breaks the build. The directives here relax that rule, and we can tighten it back up when we eventually require a minimum version that is 1.59.0 or later.

pubnub-core/src/mock/runtime.rs
stmpjmpr: If the last line does not end in a semicolon, the return value of the expression is also the return value of the function. If you put a semicolon, it is not used as the return value for the function. In all of these cases, the linter was complaining that we were implicitly returning a value that was not being used, and wanted semicolons to make that explicit.

pubnub-core/src/subscription/subscription.rs
stmpjmpr: Correct.

pubnub-core/src/mock/runtime.rs
phairow: Awesome, was guessing it was the linter complaining. Thanks for the clarification.

pubnub-core/src/lib.rs
phairow: fair enough

PR #68, opened 2023-03-01
setup the repository

Xavrax:
Setup our rust repository with:
- basic project architecture
- validations (clippy, fmt)
- tests
- README files


.github/workflows/release/versions.json
```diff
@@ -0,0 +1,28 @@
...
+  ".pubnub.yml": [
+    {
+      "pattern": "^version: (v?(\\d+\\.?){2,}([a-zA-Z0-9-]+(\\.?\\d+)?)?)$",
+      "clearedPrefix": true,
+      "clearedSuffix": false
+    }
+  ],
+  "README.md": [
+    {
+      "pattern": "^pubnub\\s=\\s(?P<major>0|[1-9]\\d*)\\.(?P<minor>0|[1-9]\\d*)\\.(?P<patch>0|[1-9]\\d*)(?:-(?P<prerelease>(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$",
```
parfeon, line 11: But why Python styled Regexps when it should be JS?

```suggestion
      "pattern": "^pubnub\\s=\\s((\\d+\\.?){2,}([a-zA-Z0-9-]+(\\.?\\d+)?)?)$",
```

This pattern should handle:
* `1.2`
* `1.2.3.5`
* `1.2-alpha`
* `12.56.3-beta.6`

Is there any specific string which you tried to match?

parfeon: And there is more in the same file.

Xavrax: I searched regex for Rust versions and found this as a most popular option that works at https://regex101.com/.

parfeon: Ah, I shared that resource to practice with regular expression :)
Scripts aren't concerned about named groups. It will use the first captured group (event thought, there are subgroups) and replace the captured value with a new version.

.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
 *         @Xavrax @mohitpubnub @kleewho @marcin-cebo
 .github/* @parfeon @Xavrax @mohitpubnub @kleewho @marcin-cebo
-**/README.md @techwritermat @kazydek @michaljolender
```
parfeon: @Xavrax are you sure about that? What if there will be `CHANGELOG.md` which pretty often is present in repos?

Xavrax: Ahhh. you're right. I will come back to all readmes.

.github/workflows/release/versions.json
Xavrax: fixed

.github/CODEOWNERS
Xavrax: I reverted and now only Readme.md files are taken into account.

parfeon (approved):
LGTM!

PR #69, opened 2023-03-06
contract tests initial setup

mohitpubnub:
* spike: cucumber-rs integration.
* sample step definition.
* hooks implementation for contract server initialization.


mohitpubnub:
@Xavrax  Have a look on defining how this workflow will trigger it.
WIP: understanding of managing test suites via cargo.

Cargo.toml
```diff
@@ -1,5 +1,15 @@
 [workspace]
 
+[dev-dependencies]
```
Xavrax, line 3: Move it the `crates/contract-tests/Cargo.toml`

tests/contract.rs
```diff
@@ -0,0 +1,63 @@
+use cucumber::{ given, when, then, World };
```
Xavrax, line 1: We have contract tests crate - move theses files there. 
`crates/contract-tests/tests/...`

Xavrax (changes requested):

tests/contract.rs
mohitpubnub: moved

Xavrax (approved):
if we're fine with the that licenses it LGTM

crates/contract-tests/tests/contract.rs
```diff
@@ -0,0 +1,67 @@
...
+                        .find(|&t| t.starts_with("contract="))
+                        .unwrap();
+                    let splitted_values: Vec<&str> = tag.split("=").collect();
+                    if !splitted_values[1].is_empty() {
+                        let script_name = splitted_values[1];
+                        init_server(script_name.to_string()).await;
+                    }
+                }
+            })
+        })
```
Xavrax, line 64:
```suggestion
    PubnubWorld::cucumber()
        .before(|_feature, _rule, scenario, _world| {
            futures::FutureExt::boxed(async move {
                if scenario
                    .tags
                    .iter()
                    .find(|&t| t.starts_with("contract="))
                    .is_some()
                {
                    let tag = _scenario
                        .tags
                        .iter()
                        .find(|&t| t.starts_with("contract="))
                        .unwrap();
                    let splitted_values: Vec<&str> = tag.split("=").collect();
                    if !splitted_values[1].is_empty() {
                        let script_name = splitted_values[1];
                        init_server(script_name.to_string()).await;
                    }
                }
            })
        })
```
Underscore is used for not used variables.

PR #70, opened 2023-03-07
deny warnings in code

Xavrax:
This PR contains:
 - denying warnings in code
 - fixing all existing warnings in current code.
 - removing `--verbose` argument because it seems to make error messages harder to understand


mohitpubnub (approved):

mohitpubnub:
regarding `init_server()` I keep it like that as we go further with some real tests.. We may tweak that init process.

Xavrax:
> regarding init_server() I keep it like that as we go further with some real tests.. We may tweak that init process.
Yeah! You're right.

Let's keep it as it is and we can incrementally tweak that during the implementation. ;d

PR #72, opened 2023-03-10
Initial middleware and transport definitions

kleewho:


crates/pubnub_core/src/middleware.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::transport_response::TransportResponse;
+
+#[derive(PartialEq)]
+pub enum MiddlewareStatus {
+    Completed(TransportRequest, TransportResponse),
+    Continue(TransportRequest, TransportResponse),
+}
+
+pub trait Middleware {
+    fn process(&self, middleware_status: MiddlewareStatus) -> Result<MiddlewareStatus, ()>;
```
mohitpubnub, line 13: I think a we will need a mechanism or a placeholder to decide whether to execute `process` or not.
Where the certain conditions(expression) can be put. And Middleware would perform check on it, And If the incoming request satisfies that condition then only it will `process` it.

parfeon: The idea was, that during `process` middleware will decide whether further processing (and passing to next middlewares) required or it should break the processing chain and send results / error back.

Xavrax: Yeah. The `MiddlewareStatus` is the indicator what middleware should do. 
But I want to point that I'm a little bit worried about that middleware approach for the Rust SDK. As we work on Friday, there is many things to do, many issues that can be appear and many shady places that makes me comfortable to say that it's not the best idea to go into it.

Xavrax (approved):
LGTM for now

PR #73, opened 2023-03-13
Simple, very simple publish is working

kleewho:
That's a WIP


crates/pubnub/src/pubnub_client.rs
```diff
@@ -0,0 +1,4 @@
+//! TODO: Add documentation
+
+/// TODO: Add documentation
```
Xavrax, line 3: We have to remember to provide the description for the modules in the meanwhile. ;d

Xavrax (approved):

mohitpubnub (approved):
Looks fine for `very simple` scope

PR #74, opened 2023-03-14
refactor: logging in transport

mohitpubnub:


crates/pubnub_transport_reqwest/src/reqwest.rs
```diff
@@ -13,9 +14,12 @@ struct TransportReqwest {
 #[async_trait::async_trait]
 impl Transport for TransportReqwest {
     async fn send(&self, request: TransportRequest) -> Result<TransportResponse, PubNubError> {
+        let path = prepare_path(&request.path, &request.query_parameters);
+        info!("{} {}", request.method, path);
+        let requet_url = format!("{}{}", &self.hostname, path);
```
parfeon, line 19: @mohitpubnub should we print `hostname` somewhere in logs?

crates/pubnub_transport_reqwest/src/reqwest.rs
```diff
@@ -13,9 +14,12 @@ struct TransportReqwest {
...
         let result = match request.method {
-            TransportMethod::Get => self.send_via_get_method(request).await,
-            TransportMethod::Post => self.send_via_post_method(request).await,
+            TransportMethod::Get => self.send_via_get_method(request, requet_url).await,
+            TransportMethod::Post => self.send_via_post_method(request, requet_url).await,
```
parfeon, line 22:
```suggestion
            TransportMethod::Get => self.send_via_get_method(request, request_url).await,
            TransportMethod::Post => self.send_via_post_method(request, request_url).await,
```

crates/pubnub_transport_reqwest/src/reqwest.rs
parfeon, line 19:
```suggestion
        let request_url = format!("{}{}", &self.hostname, path);
```

parfeon: @mohitpubnub should we print `hostname` somewhere in logs?

mohitpubnub: YESS!! That's what My suggestion was during discussion.

crates/pubnub_transport_reqwest/src/reqwest.rs
```diff
@@ -13,9 +14,12 @@ struct TransportReqwest {
 #[async_trait::async_trait]
 impl Transport for TransportReqwest {
     async fn send(&self, request: TransportRequest) -> Result<TransportResponse, PubNubError> {
+        let path = prepare_path(&request.path, &request.query_parameters);
+        let request_url = format!("{}{}", &self.hostname, path);
```
Xavrax, line 18: Maybe we can move making the request_url into the `prapare_path` function and rename it to `prepare_url`.

crates/pubnub_transport_reqwest/src/reqwest.rs
```diff
@@ -81,6 +80,9 @@ fn prepare_path(path: String, query_params: HashMap<String, String>) -> String {
 
 #[cfg(test)]
 mod should {
+    fn init() {
+        env_logger::init();
+    }
```
Xavrax, line 85: tbh do we need that logging in the tests?
Unit tests probably will exactly tell you what's wrong so imo we should make small UT to avoiding confusion. 

It would be useful in the contract tests when we will check every part of the SDK if that works as expected.

kleewho (approved):

PR #80, opened 2023-03-16
test: acceptance-tests

mohitpubnub:
* update run-tests

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -17,24 +17,38 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-      - name: Checkout actions
+      - name: Run tests
+        run: |
+          cargo test --lib transport
+      - name: Cancel workflow runs for commit on error
+        if: failure()
+        uses: ./.github/.release/actions/actions/utils/fast-jobs-failure
+  acceptance-tests:
+    name: Acceptance tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout project
+        uses: actions/checkout@v3
+      - name: Checkout mock-server action
         uses: actions/checkout@v3
         with:
           repository: pubnub/client-engineering-deployment-tools
           ref: v1
           token: ${{ secrets.GH_TOKEN }}
           path: .github/.release/actions
-      # TODO: setup features for contract tests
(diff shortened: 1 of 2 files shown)


.github/workflows/run-tests.yml
```diff
@@ -17,24 +17,38 @@ jobs:
...
+        with:
+          token: ${{ secrets.GH_TOKEN }}
+          features-path: tests/features
+      - name: Run acceptance tests
         run: |
-          cargo test 
-      - name: Cancel workflow runs for commit on error
-        if: failure()
-        uses: ./.github/.release/actions/actions/utils/fast-jobs-failure
+          cargo test --test contract
```
Xavrax, line 46: Shouldn't we have test modules e.g. `publish`, `subscribe`, etc. instead of just `contract`?

Xavrax (commented):
Imo it looks good.

.github/workflows/run-tests.yml
mohitpubnub: Exploring more on this.
As per My limited understanding how cargo test works - We need one entry point file with same name. And for cucumber settings one file starts execution of all tests. We can filter it out via tags later on. But the entry point/ file.rs is one.

.github/workflows/run-tests.yml
```diff
@@ -12,14 +12,14 @@ defaults:
...
+    name: Unit tests
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-      - name: Run tests
+      - name: Run unit tests
         run: |
-          cargo test --lib transport
+          cargo test --lib
```
mohitpubnub, line 22: @Xavrax   this line works as per our requirement. Isn't it!?

kleewho (approved):

.github/workflows/run-tests.yml
Xavrax: Does it omits the `tests/` directory ;o?

Xavrax (approved):
LGTM

mohitpubnub:
> Does it omits the tests/ directory ;o? 

Yes It is.  [here](https://github.com/pubnub/rust/actions/runs/4464757542/jobs/7841195096#step:3:363)
In this action it ran this 8 tests only
image

PR #83, opened 2023-03-20
Handle meta query param in publish

kleewho:


src/dx/publish.rs
```diff
@@ -83,6 +84,19 @@ fn bool_to_numeric(value: bool) -> String {
...
 
+fn serialize_meta(meta: &HashMap<String, String>) -> String {
+    let mut result = String::new();
+    result.push_str("{");
+    meta.iter().for_each(|k| {
+        result.push_str(format!("\"{}\":\"{}\",", k.0.as_str(), k.1.as_str()).as_str());
+    });
+    if result.chars().last().unwrap() == ',' {
+        result.remove(result.len() - 1);
+    }
```
Xavrax, line 95: Avoid using `unwrap()` in production code as much as possible!
```suggestion
result
        .chars()
        .last()
        .filter(|c| c == &',')
        .is_some()
        .then(|| result.remove(result.len() - 1));
```

Xavrax (changes requested):

src/dx/publish.rs
kleewho: Got better one for that :P

Xavrax (approved):

PR #91, opened 2023-03-22
phoenix-dev/publish-test

mohitpubnub:
* fixes visibility issues which is needed for PubNub configuration and transport module.
* Adding encoding at required places. message content and query params
* fix for unit tests accordingly
* some docs requirements are as todo :)


src/dx/pubnub_client.rs
```diff
@@ -92,18 +92,18 @@ where
...
+    pub instance_id: Option<String>,
 
     /// Sequence number for the publish requests
     #[builder(default = "1")]
-    pub(crate) next_seqn: u16,
+    pub next_seqn: u16,
 
     /// Configuration
-    pub(crate) config: PubNubConfig,
+    pub config: PubNubConfig,
```
Xavrax, line 106: Are you sure that these fields should be public?

src/dx/pubnub_client.rs
```diff
@@ -184,16 +184,16 @@ where
...
-    pub(crate) user_id: String,
+    pub user_id: String,
 
     /// Publish key
-    pub(crate) publish_key: Option<String>,
+    pub publish_key: Option<String>,
 
     /// Secret key
-    pub(crate) secret_key: Option<String>,
+    pub secret_key: Option<String>,
```
Xavrax, line 196: as above

src/transport/middleware.rs
```diff
@@ -9,9 +9,12 @@ pub struct PubNubMiddleware<T>
...
 {
-    pub(crate) transport: T,
-    pub(crate) instance_id: Option<String>,
-    pub(crate) user_id: String,
+    /// TODO: [docs]
+    pub transport: T,
+    /// TODO: [docs]
+    pub instance_id: Option<String>,
+    /// TODO: [docs]
+    pub user_id: String,
```
Xavrax, line 17: same here

tests/contract.rs
```diff
@@ -1,34 +1,61 @@
...
+                user_id: "user_id".to_string(),
+                instance_id: None,
+            },
+            next_seqn: 1,
+            instance_id: Some("instance".to_owned()),
+            config: PubNubConfig {
+                publish_key: Some(self.keyset.pubkey.to_owned()),
+                subscribe_key: self.keyset.subkey.to_owned(),
+                user_id: "some".to_owned(),
+                secret_key: Some("some".to_owned()),
```
Xavrax, line 49: Why not using the dedicated builder for that?

tests/contract.rs
```diff
@@ -38,9 +65,128 @@ async fn init_server(script: String) -> Result<String, Box<dyn std::error::Error
...
+    } else if param.eq("store") {
+        let store: bool = param_value != "0";
+        world.last_result = world
+            .get_pub_nub()
+            .publish_message(message)
+            .channel(channel)
+            .store(store)
+            .execute()
+            .await;
+    }
```
Xavrax, line 169: Suggestion: Use `match` instead of `else if` syntax. 
It's more readable and it's easier to add another branches.
```suggestion
    match param.as_str() {
        "meta" => {
            let meta_map: HashMap<String, String> =
                serde_json::from_str(param_value.as_str()).unwrap();
            world.last_result = world
                .get_pub_nub()
                .publish_message(message)
                .channel(channel)
                .meta(meta_map)
                .execute()
                .await;
        }
        "store" => {
            let store: bool = param_value != "0";
            world.last_result = world
                .get_pub_nub()
                .publish_message(message)
                .channel(channel)
                .store(store)
                .execute()
                .await;
        }
        _ => { /* do nothing */ }
    }
```

Xavrax (commented):
Proposal:
Mark the contract tests with the `ignored` attribute. 

It will skip this tests for our internal development (and e.g. users that want to play with our repository) but add the `cargo test --ignored` step to the CI file.

It will keep run the contract tests on CI but won't force us to run these tests during the development. 
Additionally, `cargo test` doesn't run only the unit and the integration tests but also the docs ones. Adding that tests to our suite without any attributes will be our pain in the ass. 

As an alternative we can hide it behind the `contract-tests` feature. 

Whatever we choose I am against adding this tests for every `cargo t` run. ;/
What do you think guys? @mohitpubnub @kleewho @parfeon

Xavrax (approved):
LGTM* <3

*as discussed: we have to hide that contract tests behind the flag.

PR #92, opened 2023-03-22
reference docs

Xavrax:
I removed every `TODO: <docs related>` comments. Additionally, all the documentation warnings has been fixed. 
There is also a little change in the docs and modules structure but everything has been adjusted so everything compiles and all tests are passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Feature list:
 :warning: We are under the development! ** Links below not work ** :warning:
 
 * [API reference for Rust](https://www.pubnub.com/docs/sdks/rust)
-* [Rust docs](https://docs.rs/pubnub/latest/pubnub/]
+* [Rust docs](<https://www.docs.rs/pubnub/latest/pubnub>]
 
 ## Support
 
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Feature list:
 
 ## Documentation
 
-:warning: We are under the development! ** Links below not works ** :warning:
+:warning: We are under the development! ** Links below not work ** :warning:
 
 * [API reference for Rust](https://www.pubnub.com/docs/sdks/rust)
 * [Rust docs](https://docs.rs/pubnub/latest/pubnub/]
--- a/README.md
+++ b/README.md
@@ -89,7 +89,10 @@ Feature list:
 
 ## Documentation
 
-* [API reference for Rust](TODO)
+:warning: We are under the development! ** Links below not works ** :warning:
+
+* [API reference for Rust](https://www.pubnub.com/docs/sdks/rust)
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -89,7 +89,10 @@ Feature list:
 
 ## Documentation
 
-* [API reference for Rust](TODO)
+:warning: We are under the development! **Links below not work** :warning:
```
kleewho, line 92:
```suggestion
:warning: We are under the development! **Links below do not work** :warning:
```

src/dx/publish.rs
```diff
@@ -1,18 +1,59 @@
...
+/// This struct is used to publish a message to a channel. It is used by the [`publish_message`] method of the [`PubNubClient`].
+/// The [`publish_message`] method is used to publish a message to a channel.
+/// The [`PublishMessageBuilder`] is used to build the request to be sent to the [`PubNub`] network.
+///
+/// # Examples
+/// ```rust
+/// # use pubnub::{PubNubClientBuilder, Keyset};
+///
+/// # #[tokio::main]
+/// # async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
kleewho, line 33: What does `#` mean? Why it's not on line 34 nor on lines 44+?

src/dx/publish.rs
```diff
@@ -1,18 +1,59 @@
...
 
-/// TODO: Add documentation
-pub type MessageType = String;
-
-/// TODO: Add documentation
+/// The [`PublishMessageBuilder`] is used to publish a message to a channel.
+///
+/// This struct is used to publish a message to a channel. It is used by the [`publish_message`] method of the [`PubNubClient`].
+/// The [`publish_message`] method is used to publish a message to a channel.
+/// The [`PublishMessageBuilder`] is used to build the request to be sent to the [`PubNub`] network.
```
kleewho, line 26: I think the same information is repeated several times. I'm not sure if I get it correctly, but does it mean that this piece of code is used to `publish a message to a channel`? I think it would be good to reword this. Maybe someone from docs can help us out

kleewho (approved):
Good iteration. We might need to revisit some of the places to reduce the repetition maybe, but I'm fine with how it is

src/dx/publish.rs
Xavrax: Yeah. This documentation is almost fully generated by github copilot. 
Many things may be repeated but overall it gives the right descriptions. 

I will work together with @techwritermat to increase quality of the descriptions.

src/dx/publish.rs
Xavrax: It hides that lines from the documentation (to not make a confusion) but these examples are treated as a happy path tests by `cargo test` command.

PR #95, opened 2023-03-24
Send headers with reqwest

kleewho:


src/dx/publish.rs
```diff
@@ -245,7 +246,7 @@ where
                 method: TransportMethod::Post,
                 query_parameters: query_params,
                 body: Some(m_vec),
-                ..Default::default()
+                headers: [(CONTENT_TYPE.into(), APPLICATION_JSON.into())].into(),
```
Xavrax, line 249: Do we want to have this as a default for every request?

src/transport/reqwest.rs
```diff
@@ -58,11 +59,18 @@ impl Transport for TransportReqwest {
...
+            TransportMethod::Get => self.prepare_get_method(request, request_url),
+            TransportMethod::Post => self.prepare_post_method(request, request_url),
         }?;
 
+        let result = builder
+            .headers(headers)
+            .send()
+            .await
+            .map_err(|e| TransportError(e.to_string()))?;
+
```
Xavrax, line 73: love it

src/dx/publish.rs
kleewho: I think so

Xavrax: You can reimplement `Default` trait by hand. 
It will be very handy to have `Default` implemented to as many our structs as possible according to API guidelines and this headers sounds like that.

kleewho: Oh... wait. For every one with payload, therefore probably not a deafult

Xavrax: Ahhh. You mean messages without any payload?

kleewho: Yes. I mean there's no content type if there's no content. It makes sense to set a content type only when we actually have a content

Xavrax (approved):

PR #97, opened 2023-03-28
Add signature when secret key is present

kleewho:
When publishing with POST signature need to be a v1 signature. Every other call should be signed with v2 signature. In case of providing secret key but not providing publish the initialization of client will result in error.

### DIFF
--- a/src/dx/pubnub_client.rs
+++ b/src/dx/pubnub_client.rs
@@ -8,6 +8,8 @@
 //! [`PubNub API`]: https://www.pubnub.com/docs
 //! [`pubnub`]: ../index.html
 
+use crate::core::PubNubError::ClientInitializationError;
+use crate::transport::middleware::SignatureKeySet;
 use crate::{core::PubNubError, core::Transport, transport::middleware::PubNubMiddleware};
 use derive_builder::Builder;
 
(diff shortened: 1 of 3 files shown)


src/dx/pubnub_client.rs
```diff
@@ -172,18 +174,21 @@ where
...
+                        transport: pre_build.transport,
+                        // TODO: String -> Cow<'static, str>
+                        instance_id: pre_build.instance_id.clone(),
+                        user_id: pre_build.config.user_id.clone(),
+                        signature_keys: pre_build.config.clone().signature_key_set()?,
+                    },
+                    instance_id: pre_build.instance_id,
+                    next_seqn: pre_build.next_seqn,
+                    config: pre_build.config,
+                })
```
Xavrax, line 190: Some many clones...

@parfeon I think that `Cow` will be needed quicker than I expected... ;/

[no action needed]

src/transport/middleware.rs
```diff
@@ -28,6 +34,71 @@ where
...
+        let mut signature_input = String::new();
+        signature_input.push_str(self.subscribe_key.as_str());
+        signature_input.push('\n');
+        signature_input.push_str(self.publish_key.as_str());
+        signature_input.push('\n');
+        signature_input.push_str(req.path.as_str());
+        signature_input.push('\n');
+        signature_input
+            .push_str(SignatureKeySet::handle_query_params(&req.query_parameters).as_str());
+        signature_input
```
Xavrax, line 66:
```suggestion
        format!(
            "{}\n{}\n{}\n{}",
            self.subscribe_key,
            self.publish_key,
            req.path,
            SignatureKeySet::handle_query_params(&req.query_parameters)
        )
```

src/transport/middleware.rs
```diff
@@ -28,6 +34,71 @@ where
...
+        signature_input.push_str(req.method.to_string().to_ascii_uppercase().as_str());
+        signature_input.push('\n');
+        signature_input.push_str(self.publish_key.as_str());
+        signature_input.push('\n');
+        signature_input.push_str(req.path.as_str());
+        signature_input.push('\n');
+        signature_input
+            .push_str(SignatureKeySet::handle_query_params(&req.query_parameters).as_str());
+        signature_input.push('\n');
+        signature_input
```
Xavrax, line 80:
```suggestion
        format!(
            "{}\n{}\n{}\n{}\n",
            req.method.to_string().to_ascii_uppercase(),
            self.publish_key,
            req.path,
            SignatureKeySet::handle_query_params(&req.query_parameters)
        )
```

Xavrax (approved):
Some suggestions but overall LGTM

src/dx/pubnub_client.rs
kleewho: Yes. I managed to avoid cloning the body at least, and thought that cloning when creating an instance is acceptable

PR #99, opened 2023-03-29
add deserialisation

Xavrax:


mohitpubnub:
@Xavrax  
I've encountered this:
When publish fails we get `PublishError`
e.g I got response - 
image

Which has reason
But  in our acceptance test we also need additional check for status code 
> the error status code is 400

I see that we have status from transport. But it seems somehow not propagated till `PublishError`

mohitpubnub (approved):
works for me to get standard success/failure response and content for publish!!!!

kleewho:
Cool!

PR #100, opened 2023-03-31
publish consumer contract test

mohitpubnub:
* steps definition for `publish` feature


tests/contract_test.rs
```diff
@@ -15,6 +15,7 @@ async fn init_server(script: String) -> Result<String, Box<dyn std::error::Error
 async fn main() {
     env_logger::builder().try_init().unwrap();
     PubNubWorld::cucumber()
+        .max_concurrent_scenarios(1)
```
Xavrax, line 18: I know that this is added because of Tomato's behaviour but imo some kind of comment why it cannot be run in concurrent style would be great.

Xavrax (approved):
LGTM

PR #108, opened 2023-04-04
test report

mohitpubnub:


.github/workflows/run-tests.yml
```diff
@@ -47,6 +47,11 @@ jobs:
       - name: Run acceptance tests
         run: |
           cargo test --features contract_test --test contract_test -- -t "not @na=rust and @featureSet=publish"
+      - name: Expose acceptance tests reports
+        uses: actions/upload-artifact@v3
+        with:
+          name: acceptance-test-reports
+          path: tests/report.xml
```
parfeon, line 54: Reports stored in `tests` better to have some different names, which will make it possible to identify whether it is from regular or `beta` tests.

```suggestion
          path: tests/report.xml
          retention-days: 7
```

parfeon (approved):
Small change would be great, but as for the rest, it looks good.

_Small note, that these changes will only upload results to artefacts, actual upload happens during release._

.github/workflows/run-tests.yml
mohitpubnub: Was thinking of making beta test result report as tests/report-beta.xml
Does this make sense ?
Or We can make report-required.xml ?

parfeon: `tests/report.xml` and `tests/report-beta.xml` are good names ;)

PR #114, opened 2023-04-11
Clen 1170 parse token

kleewho:


Xavrax (approved):
LGTM

src/dx/parse_token.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(untagged)]
+pub enum MetaValue {
+    String(String),
+    Integer(i64),
+    Float(f64),
+    Bool(bool),
+    /// It's always [`None`]
+    Null(Option<String>),
```
parfeon, line 96: @kleewho what will happen, if `null` will be in the payload? 
With `meta` in grant token I had to do this:

```rust
pub enum GrantTokenMetaValue {
    String(String),
    Integer(i64),
    Float(f64),
    Boolean(bool),
    Null,
}

#[cfg(feature = "serde")]
impl serde::Serialize for GrantTokenMetaValue {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        match self {
            GrantTokenMetaValue::String(val) => serializer.serialize_str(val),
            GrantTokenMetaValue::Integer(val) => serializer.serialize_i64(*val),
            GrantTokenMetaValue::Float(val) => serializer.serialize_f64(*val),
            GrantTokenMetaValue::Boolean(val) => serializer.serialize_bool(*val),
            GrantTokenMetaValue::Null => serializer.serialize_unit(),
        }
    }
}
```

This way, it properly encodes `GrantTokenMetaValue::Null` into `null` (JSON `null`, not string `"null"`).

kleewho: This will happen `("other".into(), Null(None)),`. It's in the test

PR #117, opened 2023-04-14
refactor/string use

mohitpubnub:


src/dx/pubnub_client.rs
```diff
@@ -227,9 +227,8 @@ where
                 Ok(PubNubClientRef {
                     transport: PubNubMiddleware {
                         transport: pre_build.transport,
-                        // TODO: String -> Cow<'static, str>
-                        instance_id: pre_build.instance_id.clone(),
-                        user_id: pre_build.config.user_id.clone(),
+                        instance_id: Arc::new(pre_build.instance_id.clone()),
```
mohitpubnub, line 230: @Xavrax  I think compiler doesn't like this way of increasing reference count for Option string type

mohitpubnub:
point out the place where we need refactoring. we may revisit this as the code base grow over the time.
it's possible that some places overlooked.

src/dx/pubnub_client.rs
mohitpubnub: wrapping in `Some` looks relevant but still it was mismatch of type.

parfeon: Can I ask why `Arc`?

mohitpubnub: this field in middleware should get reference (not clone)
since this instance is getting created and returned by build method. If we borrow it here, it's against rust's rules - it says that we can not returned value partially owned(`instance_id` field) by this function.

So reference counted string seemed proper choice here. And in async environment we can not use `Rc` so that's why it's `Arc`

Mateusz can correct me if it's not correct reason.;-)

parfeon: It is probably one of the use cases. I know only that `Arc` is good for situations, where shared resources (allocated on heap) used from multiple places and should be valid as long as it in use (even after leaving scope, where it has been allocated).

mohitpubnub: Yes, we want to share this property value between middleware and client. at least that was intention

parfeon: `Arc` mostly provide protection against deallocation. I would say, that in the current situation, as long as the instance from which we borrow this `String` is “alive”, there is no need for resource deallocation protection. Also `Arc` adds some performance overhead compared to regular memory access.

mohitpubnub: agree!
the problem what I experienced in this scenario  was that we are returning value which is referencing data in builder!! So when I had defined `'client` lifetime. It was still not able to guarantee it.

src/transport/middleware.rs
```diff
@@ -107,11 +108,11 @@ where
         req.query_parameters
             .insert("pnsdk".into(), format!("{}/{}", SDK_ID, VERSION));
         req.query_parameters
-            .insert("uuid".into(), self.user_id.clone());
+            .insert("uuid".into(), self.user_id.to_string());
 
         if let Some(instance_id) = &self.instance_id {
             req.query_parameters
-                .insert("instanceid".into(), instance_id.clone());
+                .insert("instanceid".into(), instance_id.to_string());
```
kleewho, line 115: Isn't `to_string` just a `clone` in new clothes?

kleewho (commented):
I think we should talk about this

src/transport/middleware.rs
mohitpubnub: this is for conversion so `into()` looks better.

also since this is in send() method.. this will be short lived.

src/transport/reqwest.rs
```diff
@@ -118,6 +119,11 @@ impl TransportReqwest {
         Self::default()
     }
 
+    /// Use this method to assign custom hostname
+    pub fn set_hostname(&mut self, hostname: &str) {
+        self.hostname = hostname.to_string();
+    }
+
```
Xavrax, line 126: Do we need that setter here?
`hostname` is now public so it can be changed any time. 

If you want to provide a easy to use setter for many types of data - change the param type to `S: Into<String>`

mohitpubnub: yeah! we can live without setter method here..

Xavrax: I think yes - but if you want that method. 
S: Into<String>
is also a valid way to me.

src/dx/pubnub_client.rs
```diff
@@ -227,9 +227,8 @@ where
...
+                        user_id: Arc::new(pre_build.config.user_id.clone()),
                         signature_keys: pre_build.config.clone().signature_key_set()?,
                     },
                     instance_id: pre_build.instance_id,
```
Xavrax, line 234: But wait. `instance_id` is still not a `arc`.
Both `instance_id` fields have to be in `arc` to make a sense.

Xavrax (changes requested):

src/dx/pubnub_client.rs
mohitpubnub: So at `PubNubClientRef` we should make`instance_id` as `Arc<Option<String>>` right?

mohitpubnub:
@Xavrax can you check out last commit.  Will default needed there.

src/dx/pubnub_client.rs
Xavrax: Same for `user_id`. 

Both `instance_id` and `user_id` should be under the Arc for sharing the resource. 
My bad that I haven't noticed that earlier.

mohitpubnub: ✅

mohitpubnub: @Xavrax 
I think `Arc::clone(&pre_build.instance_id)` looks more appropriate here. I've updated it but somehow I don't know why github ui is not showing it previous change outdated .

src/dx/pubnub_client.rs
```diff
@@ -131,8 +131,8 @@ where
     pub(crate) transport: T,
 
     /// Instance ID
-    #[builder(setter(strip_option), default = "None")]
-    pub(crate) instance_id: Option<String>,
+    #[builder(setter(strip_option), default = "Arc::new(Some(\"None\".to_string()))")]
+    pub(crate) instance_id: Arc<Option<String>>,
```
Xavrax, line 135: Does it mean that user has to provide the the ready to use Arc with the instance id?

I think that it will be better to add user ability to pass the String to the setter and wrap it to the arc on our own. 

Here [reference](https://docs.rs/derive_builder/latest/derive_builder/#completely-custom-fields-in-the-builder).

parfeon: `Some` of `None`? Does rust checker say something about this default: `Arc::new(None)`?

mohitpubnub: `Arc::new(None)` is valid

mohitpubnub: I think now it will be convenient for user to provide value without worrying about Arc

src/dx/pubnub_client.rs
```diff
@@ -131,8 +131,11 @@ where
     pub(crate) transport: T,
 
     /// Instance ID
-    #[builder(setter(strip_option), default = "None")]
-    pub(crate) instance_id: Option<String>,
+    #[builder(field(
+        type = "&'static str",
+        build = "Arc::new(Some(self.instance_id.to_string()))"
+    ))]
```
Xavrax, line 137: Now client cannot use Strings for setting this field. 
You can allow that by:
```suggestion
#[builder(
        setter(into),
        field(type = "String", build = "Arc::new(Some(self.instance_id))")
    )]
```

Xavrax (approved):

PR #118, opened 2023-04-14
Clen 1162

kleewho:
std, no_std, std, no_std


Xavrax:
As this PR grows and grows - I will probably split it into several ones.

PR #119, opened 2023-04-14
Add PAMv3

parfeon:
feat(pam): add grant token feature

Add grant token feature implementation.

feat(error): add service error parsing

Add variadic set of error formats parsing and their presentation with new `PubNubError` variant
`PubNubError::API { status, error, message, affected_channels, affected_channel_groups }`.

feat(pam): add token revoke feature

feat(pam): add client configuration validation

Add `PubNubClient` configuration validation to ensure that valid `subscribe_key` and `secret_key`
is set.

test(pam): add unit tests

Add unit tests for `grant_token` and `revoke_token` endpoints.

test(pam): move integration tests from access module

Move integration tests from `access` module to separate test module.

test(pam): add revoke and grant contract test

Add `grant_token` and `revoke_token` contract testing.

fix(parse_token): fix token padding issue

Fix issue because of which `parse_token` used wrong padding for passed base64-encoded token.

docs(parse_token): add documentation for parse_token

Add documentation for `parse_token` module.

feat(pam): add blocking exec

Add support for blocking requests in PAM endpoints.

feat(auth): add `auth_key` and `auth_token`

Add configuration options for user set `auth_key` or use `set_token` / `get_token` to work
with authentication token.

test: refactor contract test runner

Refactor runner code to make it filter properly `skip`, `beta` and `na=rust` tags.


src/dx/access/payloads.rs
```diff
@@ -0,0 +1,131 @@
...
+            groups: if !groups.is_empty() {
+                Some(groups)
+            } else {
+                None
+            },
+            uuids: if !uuids.is_empty() { Some(uuids) } else { None },
+        });
+    }
+
+    None
```
Xavrax, line 130: Proposal: You can use more "optional" syntax
```suggestion
    resources.map(|res| {
        let mut channels: HashMap<String, u8> = HashMap::new();
        let mut groups: HashMap<String, u8> = HashMap::new();
        let mut uuids: HashMap<String, u8> = HashMap::new();
        res.iter().for_each(|perm| match perm.resource_type() {
            ResourceType::Channel => {
                channels.insert(perm.id().to_string(), *perm.value());
            }
            ResourceType::ChannelGroup => {
                groups.insert(perm.id().to_string(), *perm.value());
            }
            ResourceType::UserId => {
                uuids.insert(perm.id().to_string(), *perm.value());
            }
        });

        GrantTokenResourcesPayload {
            channels: (!channels.is_empty()).then_some(channels),
            groups: (!groups.is_empty()).then_some(groups),
            uuids: (!uuids.is_empty()).then_some(uuids),
        }
    })
```

src/core/serializer.rs
```diff
@@ -0,0 +1,51 @@
+//! Serialization of Rust data structures.
```
Xavrax, line 1: If we provide the `Serializer` trait then we should use it in our `Serialize` trait.

Xavrax (commented):
Imo it would be nice to have example of usage in the `examples/` directory. 
One thing should be considered to do and one proposal with sweeter syntax. 
Rest LGTM.

src/dx/access/permissions.rs
```diff
@@ -0,0 +1,457 @@
...
+
+    #[test]
+    fn set_proper_update_permission_bits() {
+        assert_eq!(channel("test").update().value(), &0b0100_0000);
+    }
+
+    #[test]
+    fn set_proper_join_permission_bits() {
+        assert_eq!(channel("test").join().value(), &0b1000_0000);
+    }
```
Xavrax, line 432: You can use `test_case` here ;d

parfeon: @Xavrax, but it doesn't look like it can evaluate expressions. I can't use `channel("test").read().value()` as first argument.

src/core/serializer.rs
parfeon: @Xavrax do we need `Serialize` trait? For example, we don't have `Deserialize` trait.

src/dx/access/permissions.rs
Xavrax: Are you sure ;o
I though that it can handle expressions. ;/

parfeon: @Xavrax I tried, but it complains that it expected `,`, but found `.` in the chained call.

Xavrax: You can do it like this:
```
    #[test_case(channel("test").read().value() => 0b0000_0001)]
    fn set_proper_read_permission_bits(bits: &u8) -> u8 {
        bits.clone() // <- this clone is awful but it is required because our builder returns borrow.
    }
```

src/core/serializer.rs
Xavrax: For now it is not needed. 
Imagine the subscription. Then we have to deal with unknown type of data that user can provide. Then we will need that `deserialize` trait.

src/dx/access/permissions.rs
parfeon: @Xavrax nothing changes and output as I mentioned:

Screenshot 2023-04-18 at 10 45 39

src/core/serializer.rs
parfeon: @Xavrax so you suggest moving `Serializer` trait inside `Serialize` file?

src/dx/access/permissions.rs
parfeon: @Xavrax nm, it looks like there is some built-in `test_case` which can not, but if I use dependency – it starts evaluating ;)

.github/workflows/run-tests.yml
```diff
@@ -23,9 +29,9 @@ jobs:
       - name: Run doc tests
         run: |
           cargo test --doc
-      - name: Cancel workflow runs for commit on error
-        if: failure()
-        uses: ./.github/.release/actions/actions/utils/fast-jobs-failure
+#      - name: Cancel workflow runs for commit on error
+#        if: failure()
+#        uses: ./.github/.release/actions/actions/utils/fast-jobs-failure
```
Xavrax, line 34: why is this commented now? ;o

tests/access/access_steps.rs
```diff
@@ -0,0 +1,331 @@
...
+                    resources.channels.push(perm.get());
+                } else if permission == "MANAGE" {
+                    resources.channels.push(perm.manage());
+                } else if permission == "UPDATE" {
+                    resources.channels.push(perm.update());
+                } else if permission == "JOIN" {
+                    resources.channels.push(perm.join());
+                } else if permission == "DELETE" {
+                    resources.channels.push(perm.delete());
+                }
```
Xavrax, line 54:
```suggestion
                match permission.as_str() {
                    "READ" => resources.channels.push(perm.read()),
                    "WRITE" => resources.channels.push(perm.write()),
                    "GET" => resources.channels.push(perm.get()),
                    "MANAGE" => resources.channels.push(perm.manage());
                    "UPDATE" =>  resources.channels.push(perm.update()),
                    "JOIN" => resources.channels.push(perm.join()),
                    "DELETE" => resources.channels.push(perm.delete()),
                    _ => {/* do nothing */}
                }
```

tests/access/access_steps.rs
```diff
@@ -0,0 +1,331 @@
...
+use cucumber::{given, then, when};
+use pubnub::{
+    core::PubNubError,
+    dx::{access::permissions, parse_token},
+};
+use std::collections::HashMap;
+use std::ops::Not;
+
+/// Retrieve permission
+fn add_permission(
```
Xavrax, line 11: retrieve or add ;d?

tests/access/access_steps.rs
```diff
@@ -0,0 +1,331 @@
...
+            assert!(token_permission.manage);
+        } else if permission == "UPDATE" {
+            assert!(token_permission.update);
+        } else if permission == "JOIN" {
+            assert!(token_permission.join);
+        } else if permission == "DELETE" {
+            assert!(token_permission.delete);
+        } else {
+            matched = false;
+        }
```
Xavrax, line 318: same here

Xavrax (approved):
LGTM

.github/workflows/run-tests.yml
parfeon: @Xavrax right now, we didn't clone our GitHub actions to run in job. Do we really need it right now?
This action cancels other tasks and maybe at the moment of active development we would like to see all issues even if one of the checks failed.

tests/access/access_steps.rs
parfeon: Oh, nice!

parfeon: Applied in recent commit.

tests/access/access_steps.rs
parfeon: Fixed comment in recent commit.

tests/access/access_steps.rs
parfeon: Using `match` in recent commit.

Xavrax (approved):

PR #124, opened 2023-05-02
ref/minor-log

mohitpubnub:


mohitpubnub:
@Xavrax  seems like since we have a better error handling in place, not thinking of adding additional logs explicitly as too verbose log also make it hard to troubleshoot. 

Have gone through error cases and it seems out API error is giving ver precise info to figure out issue. E.g:
image

src/transport/reqwest.rs
```diff
@@ -56,7 +56,7 @@ pub struct TransportReqwest {
 impl Transport for TransportReqwest {
     async fn send(&self, request: TransportRequest) -> Result<TransportResponse, PubNubError> {
         let request_url = prepare_url(&self.hostname, &request.path, &request.query_parameters);
-        info!("{}", request_url);
+        info!("{} {:?} {}", request.method, request.headers, request_url);
```
Xavrax, line 59: Imo it would be nice to add some more information about where is that log generated: e.g. 
```suggestion
        info!("Sending data to pubnub: {} {:?} {}", request.method, request.headers, request_url);
```

Xavrax (commented):
Nice practice is to log our config at the beginning of the process. 
I know that probably we don't want to log pub/sub keys but I think that logging at least version could be nice. 
Additionally, if there will be more config options in the future - this log might be much more helpful. ;d

Xavrax:
> 

discussed at DM

src/dx/pubnub_client.rs
```diff
@@ -348,6 +349,7 @@ impl<T> PubNubClientConfigBuilder<T> {
             .map_err(|err| PubNubError::ClientInitialization(err.to_string()))
             .and_then(|pre_build| {
                 let token = Arc::new(spin::RwLock::new(String::new()));
+                debug!("Client Configuration: \n publish_key: {:?}\n subscribe_key: {}\n user_id: {}\n instance_id: {:?}", pre_build.config.publish_key, pre_build.config.subscribe_key, pre_build.config.user_id, pre_build.instance_id);
```
Xavrax, line 352: Why not just debug print the pre_build.config? ;o

Xavrax: or the whole client?

mohitpubnub: it may include `secret_key` or `auth_key` if user provides it. we should not reveal it.

mohitpubnub: and so I thought of being explicit about fields

mohitpubnub: we never ask user for those values AFAIK

Xavrax: ahh. Okay - I misunderstood that at DM.

Xavrax, line 352: imo it should be `info`. 

Many crates logs that at `info` level.

Xavrax (approved):
Please move it to `info` level and you can merge ;d

PR #130, opened 2023-05-16
Add release workflows

parfeon:
build(github-actions): add release workflows

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,20 +36,11 @@ jobs:
         with:
           # This should be the same as the one specified for on.pull_request.branches
           ref: master
-      - name: Checkout actions
-        uses: actions/checkout@v3
-        with:
-          repository: pubnub/client-engineering-deployment-tools
-          ref: v1
-          token: ${{ secrets.GH_TOKEN }}
-          path: .github/.release/actions
       - name: Prepare cargo environment
         shell: bash
-        run: cargo install cargo-readme
+        run: ./.github/workflows/release/pre-publish.sh
       - name: Publish to Crates
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Checkout actions
+        uses: actions/checkout@v3
+        with:
+          repository: pubnub/client-engineering-deployment-tools
+          ref: v1
+          token: ${{ secrets.GH_TOKEN }}
+          path: .github/.release/actions
+      - name: Prepare cargo environment
+        shell: bash
+        run: cargo install cargo-readme
```
Xavrax, line 48: We have to call `.github/workflows/release/pre-publish.sh` after this step.

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Publish to Crates
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
Xavrax, line 51: It will work without cargo login?
Or this actions contain `cargo login` command embedded into it?

parfeon: From docs: 
> --token token
API token to use when authenticating. This overrides the token stored in the credentials file (which is created by [cargo-login(1)](https://doc.rust-lang.org/cargo/commands/cargo-login.html)).
[Cargo config](https://doc.rust-lang.org/cargo/reference/config.html) environment variables can be used to override the tokens stored in the credentials file. The token for crates.io may be specified with the CARGO_REGISTRY_TOKEN environment variable. Tokens for other registries may be specified with environment variables of the form CARGO_REGISTRIES_NAME_TOKEN where NAME is the name of the registry in all capital letters.

PR #135, opened 2023-05-17
Better client dx

Xavrax:


src/dx/access/builders/mod.rs
```diff
@@ -19,15 +19,15 @@ pub use revoke::RevokeTokenRequestWithDeserializerBuilder;
 pub use revoke::{RevokeTokenRequest, RevokeTokenRequestBuilder};
 pub mod revoke;
 
+use crate::dx::pubnub_client::PubNubClientInstance;
```
parfeon, line 22: Isn't `pubnub_client` is redundant in path? I can see that `PubNubClientInstance` is re-exported in `lib.rs`.

parfeon (approved):
LGTM!

src/dx/access/builders/mod.rs
Xavrax: I'm gonna check that tomorrow's morning. ;o

Xavrax: It's not re-exported. ;o

parfeon: Oh, indeed. Should we? :)

Xavrax: Imo not. Developers should use it without PubNubMiddleware so we shouldn't reexport that. ;d

PR #139, opened 2023-05-18
Event engine implementation

parfeon:
feat(ee): event engine implementation

Provide generic event engine implementation which can operate over different states.

feat(ee): provide subscribe states

Provide a list of states which correspond to subscribe state machine.

feat(ee): provide list of subscribe state machine events

Provide a list of events which trigger transition between states in subscribe state machine.

feat(ee): define transitions

Define transition directions in response to event, depending on from current state machine state.

feat(ee): provide list of effects and invocations

Provide a list of effects invocation and effects which can be called as part of transitions.

### DIFF
--- a/src/core/event_engine/effect.rs
+++ b/src/core/event_engine/effect.rs
@@ -1,0 +1,12 @@
+pub trait Effect {
+    /// Unique effect identifier.
+    fn id(&self) -> String;
+
+    /// Run work associated with effect.
+    fn run<F>(&self, f: F)
+    where
+        F: Fn();
+
+    /// Cancel any ongoing effect's work.
+    fn cancel(&self);
+}
--- a/src/core/event_engine/effect.rs
+++ b/src/core/event_engine/effect.rs
@@ -1,4 +1,6 @@
-pub trait Effect {
+use crate::lib::alloc::string::String;
+
+pub(crate) trait Effect {
     /// Unique effect identifier.
     fn id(&self) -> String;
 
--- a/src/core/event_engine/effect.rs
+++ b/src/core/event_engine/effect.rs
@@ -1,13 +1,18 @@
-use crate::lib::alloc::string::String;
+use crate::{
+    core::event_engine::EffectInvocation,
+    lib::alloc::{string::String, vec::Vec},
+};
 
 pub(crate) trait Effect {
+    type Invocation: EffectInvocation;
+
     /// Unique effect identifier.
     fn id(&self) -> String;
 
     /// Run work associated with effect.
     fn run<F>(&self, f: F)
     where
-        F: Fn();
+        F: FnMut(Option<Vec<<Self::Invocation as EffectInvocation>::Event>>);
 
     /// Cancel any ongoing effect's work.
     fn cancel(&self);
--- a/src/core/event_engine/effect_dispatcher.rs
+++ b/src/core/event_engine/effect_dispatcher.rs
@@ -1,6 +1,6 @@
 use crate::{
     core::event_engine::{Effect, EffectHandler, EffectInvocation, Event},
-    lib::alloc::{rc::Rc, vec::Vec},
+    lib::alloc::{rc::Rc, vec, vec::Vec},
 };
 use phantom_type::PhantomType;
 use spin::rwlock::RwLock;
@@ -62,7 +62,8 @@ where
             effect.run(|| {
                 // Try remove effect from list of managed.
                 self.remove_managed_effect(&effect);
-                println!("Completed");
+
+                // Notify about effect run completion.
             });
         } else if invocation.cancelling() {
             self.cancel_effect(invocation);
--- a/src/core/event_engine/effect_dispatcher.rs
+++ b/src/core/event_engine/effect_dispatcher.rs
@@ -1,5 +1,5 @@
 use crate::{
-    core::event_engine::{Effect, EffectHandler, EffectInvocation, Event},
+    core::event_engine::{Effect, EffectHandler, EffectInvocation},
     lib::alloc::{rc::Rc, vec, vec::Vec},
 };
 use phantom_type::PhantomType;
@@ -11,7 +11,7 @@ pub(crate) struct EffectDispatcher<EH, EF, EI>
 where
     EI: EffectInvocation<Effect = EF>,
     EH: EffectHandler<EI, EF>,
-    EF: Effect,
+    EF: Effect<Invocation = EI>,
 {
     /// Effect invocation handler.
     ///
@@ -33,7 +33,7 @@ impl<EH, EF, EI> EffectDispatcher<EH, EF, EI>
 where
     EI: EffectInvocation<Effect = EF>,
     EH: EffectHandler<EI, EF>,
-    EF: Effect,
+    EF: Effect<Invocation = EI>,
 {
     /// Create new effects dispatcher.
     pub fn new(handler: EH) -> Self {
@@ -45,10 +45,9 @@ where
     }
 
     /// Dispatch effect associated with `invocation`.
-    pub fn dispatch<F, E>(&self, invocation: &EI, f: F)
+    pub fn dispatch<F>(&self, invocation: &EI, mut f: F)
     where
-        E: Event,
-        F: Fn(Vec<E>),
+        F: FnMut(Option<Vec<EI::Event>>),
     {
         if let Some(effect) = self.handler.create(invocation) {
             let effect = Rc::new(effect);
@@ -59,18 +58,20 @@ where
             }
 
             // Placeholder for effect invocation.
-            effect.run(|| {
+            effect.run(|events| {
                 // Try remove effect from list of managed.
                 self.remove_managed_effect(&effect);
 
                 // Notify about effect run completion.
+                // Placeholder for effect events processing (pass to effects handler).
+                f(events);
             });
         } else if invocation.cancelling() {
             self.cancel_effect(invocation);
-        }
 
-        // Placeholder for effect events processing (pass to effects handler).
-        f(vec![]);
+            // Placeholder for effect events processing (pass to effects handler).
+            f(None);
+        }
     }
 
     /// Handle effect cancellation.
@@ -80,7 +81,7 @@ where
     fn cancel_effect(&self, invocation: &EI) {
         let mut managed = self.managed.write();
         if let Some(position) = managed.iter().position(|e| invocation.cancelling_effect(e)) {
-            managed.remove(position);
+            managed.remove(position).cancel();
         }
     }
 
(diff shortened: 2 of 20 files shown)


parfeon:
FYI: Clippy won't be happy with this because a lot of code detached (it needs to be bound with actual effect implementation and EE usage).

src/core/event_engine/effect.rs
```diff
@@ -0,0 +1,12 @@
+pub trait Effect {
```
Xavrax, line 1: Will developer use this trait? Or will it be used in the API?

Xavrax:
You can add `#[allow(dead_code)]` for the whole modules.

src/core/event_engine/effect_dispatcher.rs
```diff
@@ -0,0 +1,101 @@
...
+            let effect = Rc::new(effect);
+
+            if invocation.managed() {
+                if let Ok(mut managed) = self.managed.write() {
+                    managed.push(effect.clone());
+
+                    // Drop as early as possible to avoid deadlock with `write`
+                    // lock in `remove_managed_effect`.
+                    drop(managed);
+                }
```
Xavrax, line 58: Managed will be dropped in the end of scope anyway. ;d

src/core/event_engine/effect_dispatcher.rs
```diff
@@ -0,0 +1,101 @@
...
+        // Placeholder for effect events processing (pass to effects handler).
+        f(vec![]);
+    }
+
+    /// Handle effect cancellation.
+    ///
+    /// Effects with managed lifecycle can be cancelled by corresponding effect
+    /// invocations.
+    fn cancel_effect(&self, invocation: &EI) {
+        if let Ok(mut managed) = self.managed.write() {
```
Xavrax, line 80: same here

src/core/event_engine/effect_dispatcher.rs
```diff
@@ -0,0 +1,101 @@
...
+            if let Some(position) = managed.iter().position(|e| invocation.cancelling_effect(e)) {
+                managed.remove(position);
+            }
+
+            // Drop as early as possible to avoid deadlocks.
+            drop(managed);
```
Xavrax, line 86: and here

src/core/event_engine/effect_invocation.rs
```diff
@@ -0,0 +1,21 @@
+use crate::core::event_engine::Effect;
+
+/// Effect invocation trait.
+///
+/// Invocation is an intention to run an effect. Effect dispatcher uses intents
+/// to schedule actual effect invocation.
+pub trait EffectInvocation {
```
Xavrax, line 7: Simillar to effect: Should it be public?

src/core/event_engine/mod.rs
```diff
@@ -0,0 +1,97 @@
+//! Event Engine module
+
+use std::sync::RwLock;
```
Xavrax, line 3: You have to allign that to `no_std` requirements. I can help you with that if you need.

src/dx/subscribe/event_engine/event.rs
```diff
@@ -0,0 +1,151 @@
...
+            SubscribeEvent::HandshakeReconnectFailure { .. } => "HANDSHAKE_RECONNECT_FAILURE",
+            SubscribeEvent::HandshakeReconnectGiveUp { .. } => "HANDSHAKE_RECONNECT_GIVEUP",
+            SubscribeEvent::ReceiveSuccess { .. } => "RECEIVE_SUCCESS",
+            SubscribeEvent::ReceiveFailure { .. } => "RECEIVE_FAILURE",
+            SubscribeEvent::ReceiveReconnectSuccess { .. } => "RECEIVE_RECONNECT_SUCCESS",
+            SubscribeEvent::ReceiveReconnectFailure { .. } => "RECEIVE_RECONNECT_FAILURE",
+            SubscribeEvent::ReceiveReconnectGiveUp { .. } => "RECEIVE_RECONNECT_GIVEUP",
+            SubscribeEvent::Disconnect => "DISCONNECT",
+            SubscribeEvent::Reconnect => "RECONNECT",
+        }
```
Xavrax, line 149: I think `Display` trait is made for that usage.

parfeon: Had some discussion about this and, at least for now, decided to keep as is (there is no need to make formatted messages with `Display`, just plain strings).

src/core/event_engine/effect_dispatcher.rs
parfeon: This one gone away, when I started to use `RwLock` from spin crate instead of `std`.

src/core/event_engine/effect_dispatcher.rs
parfeon: This one gone away, when I started to use `RwLock` from spin crate instead of `std`.

src/core/event_engine/mod.rs
parfeon: Made changes, and it wasn't the only place where proper `no_std`-compatible imports used. Made changes to all places.

src/core/event_engine/effect_dispatcher.rs
```diff
@@ -0,0 +1,94 @@
+use crate::{
+    core::event_engine::{Effect, EffectHandler, EffectInvocation, Event},
+    lib::alloc::{rc::Rc, vec, vec::Vec},
+};
+use phantom_type::PhantomType;
+use spin::rwlock::RwLock;
+
+/// State machine effects dispatcher.
```
Xavrax, line 8: TYI: That's nice to have docs but you're not obligated to do them to all private stuff ;d.

src/core/event_engine/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Process transition.
+    ///
+    /// This method is responsible for transition maintenance:
+    /// * update current state
+    /// * call effects dispatcher to process effect invocation
+    fn process_transition(&self, transition: Transition<S::State, S::Invocation>) {
+        let mut writable_state = self.current_state.write();
+        *writable_state = transition.state;
+        drop(writable_state);
```
Xavrax, line 91: This pattern is also pretty common in rust community:
```suggestion
        {
            let mut writable_state = self.current_state.write();
            *writable_state = transition.state;
        }
```
I prefer that one above but take this one which is preferred by you.

src/core/event_engine/state.rs
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/// State machine state trait.
+///
+/// For transition, the state machine needs to know which effects should be
+/// dispatched during transition to target state in response to a specific
+/// event.
+///
+/// Types which are expected to be used as states should implement the trait.
+pub(crate) trait State {
```
Xavrax, line 13: Do we need that trait?

src/core/event_engine/effect_handler.rs
```diff
@@ -0,0 +1,10 @@
+use crate::core::event_engine::{Effect, EffectInvocation};
+
+pub(crate) trait EffectHandler<I, E>
```
Xavrax, line 3: Do we need that trait?

src/core/event_engine/effect_invocation.rs
```diff
@@ -0,0 +1,21 @@
+use crate::core::event_engine::Effect;
+
+/// Effect invocation trait.
+///
+/// Invocation is an intention to run an effect. Effect dispatcher uses intents
+/// to schedule actual effect invocation.
+pub(crate) trait EffectInvocation {
```
Xavrax, line 7: Do we need that trait?

Xavrax (commented):
Overall LGTM but I have some questions.

Xavrax (commented):
What thing makes me worried. There are 1.5k lines without any test ;o

Xavrax (commented):
Some question that I found during working with effects.

src/core/event_engine/effect_dispatcher.rs
parfeon: I got it. It just a matter of habit.

src/core/event_engine/state.rs
parfeon: Yup. 
This is the base trait for other types of state machine `state` (right now only _subscribe_, but there will be more).

src/core/event_engine/effect_invocation.rs
parfeon: Yup. 
This is the base trait for other types of state machine `effect invocation` (right now only _subscribe_, but there will be more).

parfeon:
> What thing makes me worried. There are 1.5k lines without any test ;o

I will work on this.

src/core/event_engine/effect_dispatcher.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /// Dispatch effect associated with `invocation`.
+    pub fn dispatch<F, E>(&self, invocation: &EI, f: F)
+    where
+        E: Event,
+        F: Fn(Vec<E>),
+    {
+        if let Some(effect) = self.handler.create(invocation) {
+            let effect = Rc::new(effect);
```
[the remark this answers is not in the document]
parfeon: > Why do we need that RC here?

I have a requirement to store managed `effect` in a vector and make it possible to remove the same effect from list after `run` completion.
I tried to filter in completion block, but borrow checker won't let me go this way. Furthermore, I used `Rc` just to make clone once to add it vector and don't need `Rc` functionality after that.

> Why do we passing the remove_managed_effect into the effect and call it inside of it? 

I can't follow this one.
The idea was that `run` function of effect has completion closure, with which we will be able to remove effect from the list of managed effects. 

> Additionally, if we need reference counter it probably won't work with our client as it needs to be thread safe.
> Use Arc instead.

`Rc` functionality used once (I mean clone) within `dispatch` function, and that is all. All other access to it through vector which stores it should be safe enough because of `RwLock`.

src/core/event_engine/mod.rs
parfeon: Made this change in recent push.

src/core/event_engine/effect_dispatcher.rs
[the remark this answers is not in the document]
Xavrax: >The idea was that run function of effect has completion closure, with which we will be able to remove effect from the list of managed effects.
This one is still confusing to me. 
How effects are allowed to managed the call?
Currently in SubscribeEffect it doesn't make any decision about it. Do you think that some effects will affect that?
We still can casually call that line after calling `run` function.

src/core/event_engine/effect_dispatcher.rs
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            // Placeholder for effect invocation.
+            effect.run(|events| {
+                // Try remove effect from list of managed.
+                self.remove_managed_effect(&effect);
+
+                // Notify about effect run completion.
+                // Placeholder for effect events processing (pass to effects handler).
+                f(events);
```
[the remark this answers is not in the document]
parfeon: We don't have implemented effects yet, but I assume that some of them would return events, which should be processed by the event engine.
It is exactly as it says – it is a _placeholder_ for functionality, which I expect to appear with implemented effects.
The function pointer can be passed to the created effect, so it will be called from it.

src/dx/subscribe/event_engine/state.rs
```diff
@@ -0,0 +1,2225 @@
...
+                .chain(invocations.unwrap_or(vec![]).into_iter())
+                .chain(state.enter().unwrap_or(vec![]).into_iter())
+                .collect(),
+            state,
+        }
+    }
+}
+
+#[cfg(test)]
+mod should {
```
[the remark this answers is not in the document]
parfeon: I made some changes, but it saved only 700 lines. Pros from change: maybe it becomes more readable and easier to understand what transition should be tested for which event.

src/core/event_engine/effect_dispatcher.rs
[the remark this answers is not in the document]
Xavrax: I'm still not sure about that. 
Anyway imo you can merge that. 
Rest of codes looks good.

Xavrax (approved):
LGTM (except that part with function)

Xavrax (approved):

PR #143, opened 2023-05-30
add reconnection effect

Xavrax:


src/dx/subscribe/event_engine/event.rs
```diff
@@ -60,7 +60,7 @@ pub(crate) enum SubscribeEvent {
     /// response from [`PubNub`] network (network issues or permissions).
     ///
     /// [`PubNub`]: https://www.pubnub.com/
-    HandshakeReconnectFailure { reason: PubNubError },
+    HandshakeReconnectFailure { reason: PubNubError, attempts: u8 },
```
parfeon, line 63: We don't need to know, how many times handshake failed, before we gave up and stopped any further attempts.

parfeon (approved):
LGTM!

PR #144, opened 2023-05-30
Add request retry policy

parfeon:
feat(retry-policy): add request retry policy

Add `RequestRetryPolicy` for delay computation between failed requests to retry.

fix(reqwest-response): fix missing headers

Fix missing `headers` in `RequestResponse` struct when used with `reqwest` crate.

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -37,6 +37,10 @@ pub mod transport_response;
 #[doc(inline)]
 pub use serialize::Serialize;
 
+#[doc(inline)]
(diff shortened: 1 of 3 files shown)


Xavrax (commented):
~~LGTM but I would prefer to use [Duration](https://doc.rust-lang.org/std/time/struct.Duration.html) instead of `u32`. ~~

Xavrax (approved):

Xavrax:
But it would be nice to use e.g. https://docs.rs/time/latest/time/struct.Duration.html

PR #147, opened 2023-06-02
add receive messages effect

Xavrax:


src/dx/subscribe/event_engine/effects/mod.rs
```diff
@@ -6,9 +6,11 @@ use crate::{
 };
 
 use super::effect_handler::HandshakeFunction;
+use super::ReceiveFunction;
```
parfeon, line 9: Should we make it look closer to the way how we use `HandshakeFunction` which used through the `effect_handler` module?

parfeon (approved):
LGTM!

PR #152, opened 2023-06-06
Use custom cargo action for release

parfeon:
build(cargo): use custom cargo action for release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,7 @@ jobs:
           ref: v1
           token: ${{ secrets.GH_TOKEN }}
           path: .github/.release/actions
-      - name: Publish to CocoaPods Trunk
+      - name: Publish to crates
         uses: ./.github/.release/actions/actions/services/crates
         with:
           token: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -36,11 +36,22 @@ jobs:
...
-        run: ./.github/workflows/release/publish.sh
+          token: ${{ secrets.GH_TOKEN }}
+      - name: Checkout actions
+        uses: actions/checkout@v3
+        with:
+          repository: pubnub/client-engineering-deployment-tools
+          ref: v1
+          token: ${{ secrets.GH_TOKEN }}
+          path: .github/.release/actions
+      - name: Publish to CocoaPods Trunk
```
Xavrax, line 47: What is the purpose of this action?

Xavrax (approved):
LGTM but I have question

.github/workflows/release.yml
parfeon: Is it ok now?

Xavrax: I didn't requested for change. I was curious what exactly was this action doing. ;d
Anyway made change is also good. ;d

Xavrax (approved):

.github/workflows/release.yml
parfeon: It is an action, which will push to crates and run all our logic with notifications and other stuff :)

PR #161, opened 2023-07-24
add custom origin example

Xavrax:


techwritermat (approved):

examples/custom_origin.rs
```diff
@@ -0,0 +1,38 @@
...
+            secret_key: None,
+        })
+        .with_user_id("user_id")
+        .build()?;
+
+    // publish to check the custom origin
+    let result = client
+        .publish_message("hello world!")
+        .channel("my_channel")
+        .r#type("text-message")
```
kleewho, line 31: Do we want to support this?

mohitpubnub (approved):

kleewho (approved):

examples/custom_origin.rs
mohitpubnub: You mean r#type right ?

Xavrax: I can remove that after we will get rid of it in general.

PR #165, opened 2023-08-26
Add `@parfeon` as code owner

parfeon:
chore(codeowners): add `@parfeon` as code owner


Xavrax (approved):
Yooo... Why weren't you here before ;o?

mohitpubnub (approved):
SDK Author was not codeowner till now

PR #169, opened 2023-08-29
where now presence API

Xavrax:


src/dx/presence/mod.rs
```diff
@@ -213,6 +213,46 @@ impl<T, D> PubNubClientInstance<T, D> {
...
+
+    ///
+    /// println!("User channels: {:?}", response);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn where_now(&self) -> WhereNowRequestBuilder<T, D> {
+        WhereNowRequestBuilder {
+            pubnub_client: Some(self.clone()),
```
parfeon, line 252: @Xavrax what do you think about default `user_id` usage here from `client` configuration?

parfeon (approved):
LGTM!

src/dx/presence/mod.rs
Xavrax: Tbh. I prefer when API is more explicit but I'm not against the default value.

parfeon: We had this behaviour with default `user_id` in other SDKs for some endpoints.

parfeon: But it is totally fine as it now ;)

Xavrax: Okay. I don't want to make it different than in other SDKs. I will change that - but to keep git history not so complicated, I will change that in the nexts PRs. (I'm planning to make `state` and `no_std` PRs before release so I can make that change during one of those)

Xavrax: fixed.

PR #170, opened 2023-08-30
Subscribe EE works with presence EE

parfeon:
feat(event_engine): subscribe EE works with presence EE

On subscription changes and restore, the subscribe event engine should inform the presence event engine about such events so `user_id` presence will be properly updated.

feat(subscribe): add missing `disconnect` and `reconnect` methods

Add method which lets users temporarily stop real-time updates processing and restore when it will be needed.

### DIFF
--- a/src/dx/subscribe/mod.rs
+++ b/src/dx/subscribe/mod.rs
@@ -110,43 +110,6 @@ where
         }
     }
 
-    /// Invalidate subscription listener.
-    ///
-    /// All previously created listeners will be invalidated and stop receiving
-    /// real-time updates from specified list of channels and groups.
-    ///
-    /// ```no_run
-    /// use futures::StreamExt;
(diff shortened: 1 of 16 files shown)


src/dx/subscribe/mod.rs
```diff
@@ -110,21 +110,203 @@ where
...
+    /// #         .stream();
+    /// client.unsubscribe_all();
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn unsubscribe_all(&self) {
+        if let Some(manager) = self.subscription.write().as_mut() {
+            manager.unregister_all();
+        }
+    }
```
Xavrax, line 148: I'm not a fan of that `unsubscribe_all` method at all. 
As we return handles to the `Subscription`s, we shouldn't give a way to invalid them from other place then itself.

Xavrax (approved):
LGTM (except 1 thing)

PR #171, opened 2023-08-30
add get_presence_state to presence API

Xavrax:


src/dx/presence/result.rs
```diff
@@ -239,6 +239,143 @@ impl TryFrom<SetStateResponseBody> for SetStateResult {
     }
 }
 
+/// The result of a get state operation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct GetStateResult {
+    /// State which has been associated for `user_id` with channel(s) or channel
+    /// group(s).
+    pub states: Vec<GetStateInfo>,
```
parfeon, line 247: @Xavrax can it be a `HashMap`? Iterating through the list of states is the only way for the user to check whether a specific channel state is set (when requested for multiple channels).

parfeon (changes requested):
I think small change required.

src/dx/presence/result.rs
Xavrax: But... We do not lose this info during using the `Vec` because inner type contains `channel` and `state` fields ;o
Imo HashMap UX would be worse because the arguments are not named and user have to undestand that key-value is channel-state.

parfeon: Is there an easy way to get from the response state for a particular channel? I don't say that the inner structure needs to be changed.

Xavrax: Yes, you can use `filter` method for `iterators` like so:
```rust
result.iter().filter(|a| a.channel == "asd".into())

parfeon: @Xavrax i knew you would say that 🤣

parfeon (approved):

PR #173, opened 2023-08-30
Release `v0.3.0`

Xavrax:
feat: subscribe API implemented

PubNub subscribe API implemented

feat: presence API implemented

PubNub presence API implemented

feat: event engine for presence and subscribe API

Event engine as a new method of handling connections for `subscribe` and `presence` methods implemented


.pubnub.yml
```diff
@@ -87,6 +87,22 @@ features:
...
+  subscribe:
+    - SUBSCRIBE-CHANNELS
+    - SUBSCRIBE-CHANNEL-GROUPS
+    - SUBSCRIBE-PRESENCE-CHANNELS
+    - SUBSCRIBE-PRESENCE-CHANNELS-GROUPS
+    - SUBSCRIBE-WITH-TIMETOKEN
+    - SUBSCRIBE-FILTER-EXPRESSION
+    - SUBSCRIBE-PUBLISHER-UUID
+    - SUBSCRIBE-PUBSUB-V2
+    - SUBSCRIBE-MESSAGE-ACTIONS-LISTENER
```
parfeon, line 99: @Xavrax if this one about separate listeners - we don't have separate listener for actions.
If it is about listener's ability to inform about those events, then in addition we have:
* `SUBSCRIBE-SIGNAL-LISTENER`
* `SUBSCRIBE-MEMBERSHIP-LISTENER`
* `SUBSCRIBE-OBJECTS-CHANNEL-LISTENER`
* `SUBSCRIBE-OBJECTS-UUID-LISTENER`
* `SUBSCRIBE-FILE-LISTENER`

.pubnub.yml
```diff
@@ -87,6 +87,22 @@ features:
...
+  presence:
+    - PRESENCE-HERE-NOW
+    - PRESENCE-WHERE-NOW
+    - PRESENCE-SET-STATE
+    - PRESENCE-GET-STATE
+    - PRESENCE-HEARTBEAT
```
parfeon, line 105: @Xavrax it is confusing because section, but I think we need this one as well: `PRESENCE-DELTAS` - our subscribe parser able to handle interval update of present users. 
@techwritermat is this flag about interval update of presence (when we receive set of joined/left and timeout)?

src/lib.rs
```diff
@@ -144,11 +144,14 @@
 //!
 //! | Feature name  | Description | Available PubNub APIs |
 //! | :------------ | :---------- | :------------- |
-//! | `full`        | Enables all non-conflicting features | Configuration, Publish, Access Manager, Parse Token |
-//! | `default`     | Enables default features: `publish`, `serde`, `reqwest`, `aescbc`, `std` | Configuration, Publish |
+//! | `full`        | Enables all non-conflicting features | Configuration, Publish, Subscribe, Access Manager, Parse Token, Presence |
+//! | `default`     | Enables default features: `publish`, `subscribe`, `serde`, `reqwest`, `aescbc`, `std` | Configuration, Publish, Subscribe |
```
parfeon, line 148: @Xavrax in `Cargo.toml` I've added `presence` as part of the default. It shouldn't be in the default list?
Also, we are missing `access` which is in `Cargo.toml` rn.

parfeon (approved):

src/lib.rs
Xavrax: I've remove presence from default (because it's not default in portal)

Xavrax: I see access in cargo `full` and lib.rs `table`

.pubnub.yml
techwritermat: DM'd you

.pubnub.yml
techwritermat: DM'd you

Xavrax:
@pubnub-release-bot release as 0.3.0

techwritermat (approved):

PR #175, opened 2023-09-21
Add crypto module

parfeon:
feat(crypto): add crypto module

Update the crypto module structure and add enhanced AES-CBC cryptor.

### DIFF
--- a/examples/crypto.rs
+++ b/examples/crypto.rs
@@ -45,7 +45,7 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
         .with_cryptor(crypto_module)
         .build()?;
 
-    // publish simple string
+    // publish encrypted string
     let result = client
         .publish_message("hello world!")
         .channel("my_channel")
--- a/examples/crypto.rs
+++ b/examples/crypto.rs
@@ -3,11 +3,23 @@ use pubnub::providers::crypto::CryptoModule;
 use pubnub::{Keyset, PubNubClientBuilder};
 use std::env;
 
+/// This example demonstrates how data can be manually encrypted or
+/// automatically as part of PubNubClient instance when publish / subscribe is
+/// used.
+///
+/// The following example consists of two parts: a manual _encryption_ and
+/// _decryption_ demonstration, and an automated _encryption_ and _decryption_
+/// demonstration, as part of a configured `PubNubClientInstance`.
+
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn snafu::Error>> {
     let source_data: Vec<u8> = "Hello world!".into();
     let use_random_iv = true;
     let cipher = "enigma";
+
+    // -----------------------------------------
+    // Manual encryption and decryption example.
+    //
 
     // Crypto module with legacy AES-CBC cryptor (with enhanced AES-CBC decrypt
     // support).
@@ -30,6 +42,10 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
     // Decrypt data created with enhanced AES-CBC crypto module.
     let decrypt_result = legacy_crypto_module.decrypt(encrypt_result.ok().unwrap())?;
     assert_eq!(decrypt_result, source_data);
+
+    // --------------------------------------------
+    // Automated encryption and decryption example.
+    //
 
     // Setup client with crypto module
     let publish_key = env::var("SDK_PUB_KEY")?;
--- a/examples/crypto.rs
+++ b/examples/crypto.rs
@@ -1,0 +1,59 @@
+use pubnub::core::CryptoProvider;
+use pubnub::providers::crypto::CryptoModule;
+use pubnub::{Keyset, PubNubClientBuilder};
+use std::env;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn snafu::Error>> {
+    let source_data: Vec<u8> = "Hello world!".into();
+    let use_random_iv = true;
+    let cipher = "enigma";
+
+    // Crypto module with legacy AES-CBC cryptor (with enhanced AES-CBC decrypt
+    // support).
+    let legacy_crypto_module = CryptoModule::new_legacy_module(cipher, use_random_iv)?;
+    let legacy_encrypt_result = legacy_crypto_module.encrypt(source_data.clone());
+
+    println!("encrypt with legacy AES-CBC result: {legacy_encrypt_result:?}");
+
+    // Crypto module with enhanced AES-CBC cryptor (with legacy AES-CBC decrypt
+    // support).
+    let crypto_module = CryptoModule::new_aes_cbc_module(cipher, use_random_iv)?;
+    let encrypt_result = crypto_module.encrypt(source_data.clone());
+
+    println!("encrypt with enhanced AES-CBC result: {encrypt_result:?}");
+
+    // Decrypt data created with legacy AES-CBC crypto module.
+    let legacy_decrypt_result = crypto_module.decrypt(legacy_encrypt_result.ok().unwrap())?;
+    assert_eq!(legacy_decrypt_result, source_data);
+
+    // Decrypt data created with enhanced AES-CBC crypto module.
+    let decrypt_result = legacy_crypto_module.decrypt(encrypt_result.ok().unwrap())?;
+    assert_eq!(decrypt_result, source_data);
+
+    // Setup client with crypto module
+    let publish_key = env::var("SDK_PUB_KEY")?;
+    let subscribe_key = env::var("SDK_SUB_KEY")?;
+
+    let client = PubNubClientBuilder::with_reqwest_transport()
+        .with_keyset(Keyset {
+            subscribe_key,
+            publish_key: Some(publish_key),
+            secret_key: None,
+        })
+        .with_user_id("user_id")
+        .with_cryptor(crypto_module)
+        .build()?;
+
+    // publish simple string
+    let result = client
+        .publish_message("hello world!")
+        .channel("my_channel")
+        .r#type("text-message")
+        .execute()
+        .await?;
+
+    println!("publish result: {:?}", result);
+
+    Ok(())
+}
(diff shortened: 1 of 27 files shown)


examples/crypto.rs
```diff
@@ -0,0 +1,59 @@
...
+        .with_keyset(Keyset {
+            subscribe_key,
+            publish_key: Some(publish_key),
+            secret_key: None,
+        })
+        .with_user_id("user_id")
+        .with_cryptor(crypto_module)
+        .build()?;
+
+    // publish simple string
```
Xavrax, line 48:
```suggestion
    // publish encrypted string
```

examples/crypto.rs
```diff
@@ -0,0 +1,59 @@
+use pubnub::core::CryptoProvider;
+use pubnub::providers::crypto::CryptoModule;
+use pubnub::{Keyset, PubNubClientBuilder};
+use std::env;
+
```
Xavrax, line 5: it would be nice to have some kind of description that user can encrypt data partially (the first part of the example) and fully automated (the second one).

src/core/crypto_provider.rs
```diff
@@ -0,0 +1,27 @@
...
+//! implement a module that can be used to configure [`PubNubClientInstance`] or
+//! for manual data encryption and decryption.
+
+use crate::{
+    core::PubNubError,
+    lib::{alloc::vec::Vec, core::fmt::Debug},
+};
+
+/// Crypto provider trait.
+pub trait CryptoProvider: Debug + Send + Sync {
```
Xavrax, line 13: Do we need that `Debug` here? Do we print any debug data for Crypto Provider?

src/core/cryptor.rs
```diff
@@ -1,59 +1,98 @@
...
+///     fn decrypt(&self, source: EncryptedData) -> Result<Vec<u8>, PubNubError> {
 ///         // Decrypt provided data here
 ///         Ok(vec![])
 ///     }
 /// }
 /// ```
-///
-/// [`dx`]: ../dx/index.html
-/// [`PubNub API`]: https://www.pubnub.com/docs
 pub trait Cryptor: Debug + Send + Sync {
```
Xavrax, line 76: as above

src/lib.rs
```diff
@@ -145,7 +147,7 @@
 //! | Feature name  | Description | Available PubNub APIs |
 //! | :------------ | :---------- | :------------- |
 //! | `full`        | Enables all non-conflicting features | Configuration, Publish, Subscribe, Access Manager, Parse Token, Presence |
-//! | `default`     | Enables default features: `publish`, `subscribe`, `serde`, `reqwest`, `aescbc`, `std` | Configuration, Publish, Subscribe |
+//! | `default`     | Enables default features: `publish`, `subscribe`, `serde`, `reqwest`, `crypto`, `std` | Configuration, Publish, Subscribe |
```
Xavrax, line 150: add `crypto` into `full` feature.

btw. do we want to have it as a default one?

src/providers/crypto/cryptors/legacy.rs
```diff
@@ -0,0 +1,275 @@
...
+        })
+    }
+
+    fn initialization_vector(&self) -> [u8; 16] {
+        if self.use_random_iv {
+            let mut random = [0u8; AES_BLOCK_SIZE];
+            getrandom::getrandom(&mut random).ok();
+            random
+        } else {
+            *b"0123456789012345"
```
Xavrax, line 70: Wow, I didn't know that it is copiable. ;o

src/providers/crypto/cryptors/legacy.rs
```diff
@@ -0,0 +1,275 @@
...
+        let data_offset = if self.use_random_iv {
+            AES_BLOCK_SIZE
+        } else {
+            0
+        };
+        let iv = if self.use_random_iv {
+            data.data[0..AES_BLOCK_SIZE].to_vec()
+        } else {
+            self.initialization_vector().to_vec()
+        };
```
Xavrax, line 154:
```suggestion
        let iv = if self.use_random_iv {
            data.data[0..AES_BLOCK_SIZE]
        } else {
            self.initialization_vector()
        }.to_vec();
```
nitpicking

Xavrax (approved):
great job!

src/core/crypto_provider.rs
parfeon: @Xavrax our `PubNubClientInstance` derive from Debug – and this is where everything began :D

src/core/cryptor.rs
parfeon: @Xavrax because `Cryptor` is part of another trait / interface which derives from `Debug` it also needs to have `Debug`.

src/lib.rs
parfeon: @Xavrax, I will add, however, regarding the issue of `default`, I do not possess a clear opinion. I would suspect that we enable everything by default, or no?

src/providers/crypto/cryptors/legacy.rs
parfeon: Yeah, my bad :/

src/core/crypto_provider.rs
Xavrax: Makes sense.

Xavrax:
my bad. I see that `data` and `initialization_vector` has different types. 
Probably they are arrays with different amount of entities. 
Maybe revert that change then.

src/lib.rs
Xavrax: I'm not sure if we should enable everything as default. 
Imo we should have as small amount default features as possible but with taking to the account dx of our customers.

src/providers/crypto/cryptors/aes_cbc.rs
```diff
@@ -0,0 +1,205 @@
...
+                details: "Cipher key is empty".into(),
+            });
+        }
+
+        Ok(Self {
+            cipher_key: Self::sha256(cipher_key),
+        })
+    }
+
+    fn initialization_vector(&self) -> [u8; 16] {
```
josh-lubliner, line 50:
```suggestion
    fn initialization_vector(&self) -> [u8; AES_BLOCK_SIZE] {
```
just for consistency

src/providers/crypto/cryptors/aes_cbc.rs
```diff
@@ -0,0 +1,205 @@
...
+        let mut random = [0u8; AES_BLOCK_SIZE];
+        getrandom::getrandom(&mut random).ok();
```
josh-lubliner, line 52: I'm actually not sure about this.  The `getrandom` crate's docs says this:

> It is assumed that the system always provides high-quality cryptographically secure random data

Is it safe to make this assumption?  On the other hand, even the `rand` crate, which explicitly says it provides cryptographically secure random number generators, says this:

> A Rust library for random number generation, featuring secure seeding via the getrandom crate

So it isn't clear that using the `rand` crate would be any better.  My best guess is that this will be fine for nearly all customers.  However, there may be customers whose software runs in an environment (e.g. embedded hardware) that does not have access to cryptographically secure seeding.  While it would be possible to allow customers to provide their own custom RNG in those cases, probably it's best that those customers just don't use this feature.

All of which is to say I am not recommending any change here.

src/providers/crypto/cryptors/aes_cbc.rs
```diff
@@ -0,0 +1,205 @@
...
+
+use aes::cipher::{block_padding::Pkcs7, BlockDecryptMut, BlockEncryptMut, KeyIvInit};
+use sha2::{Digest, Sha256};
+
+use crate::{
+    core::{Cryptor, EncryptedData, PubNubError},
+    lib::alloc::{format, string::ToString, vec, vec::Vec},
+};
+
+type Encryptor = cbc::Encryptor<aes::Aes256>;
```
josh-lubliner, line 14: The docs for the `aes` crate say:

> ⚠️ Warning: Hazmat!  This crate does not ensure ciphertexts are authentic (i.e. by using a MAC to verify ciphertext integrity), which can lead to serious vulnerabilities if used incorrectly!  To avoid this, use an AEAD mode based on AES, such as AES-GCM or AES-GCM-SIV.  USE AT YOUR OWN RISK!

Obviously this is not great.  Solutions are to use AES-GCM, which would involve changing all the SDKs, find an AES crate which does not have this vulnerability (they exist, but they haven't been audited.  At least the `aes` crate has been audited), or live with this issue.  Another possibility would be to MAC the messages ourselves, but that's probably just as hard as switching to AES-GCM, and more error-prone.

I'm not sure what the level of vulnerability is here.  We might need an actual cryptographer to weigh in.

src/providers/crypto/cryptors/aes_cbc.rs
parfeon: Yes. It looks like all suitable (`no_std`) CSPRNGs from [this](https://rust-random.github.io/book/guide-rngs.html#cryptographically-secure-pseudo-random-number-generators-csprngs) list rely on `getrandom` when we need to create an instance of random number generator by requesting create it from `entropy` – this under the hood relies on the `getrandom` feature.

Currently, it is hard to say what is the best solution here. With future updates, _maybe_ we will create some deviation (from other SDKs) in the interface to let user specify closure which will generate random data for IV (and salt when we will switch to KDF).

src/providers/crypto/cryptors/aes_cbc.rs
parfeon: Before we had were advised to use AES-GCM to protect against [Padding Oracles](https://research.nccgroup.com/2021/02/17/cryptopals-exploiting-cbc-padding-oracles/), but all _encryption_ and _decryption_ happens on the device (end to end) so there is no Oracle (server) to exploit. If there were servers, then yes, AES-GCM with authentication tags would help here, but it is just not our case.

josh-lubliner: After some consultation, I think you are basically correct.  Perhaps a future GCM-based cryptor, but not changes here for now.

parfeon:
@pubnub-release-bot release

kazydek (approved):

PR #180, opened 2023-11-07
add new PubNub license

Xavrax:
fix: Changed license type from MIT to PubNub Software Development Kit License


Xavrax:
I missed that it has been done. My bad.

PR #182, opened 2024-01-19
update reqwest features to use rustls

CodyPubNub:
Switches reqwest from using openssl to instead use rustls

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,7 @@ serde = { version = "1.0", features = ["derive"], optional = true, default-featu
 serde_json = { version = "1.0", optional = true, features = ["alloc"] ,default-features = false }
 
 # reqwest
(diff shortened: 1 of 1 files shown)


stephenlb (approved):

Xavrax (approved):
LGTM

Xavrax:
We will merge the changes when the new clippy tool rules are satisfied by the master branch.

parfeon:
We will push this with upcoming release with event listeners.

parfeon:
@CodyPubNub looks like our linters won't pass this PR.
* `ring` which is used by `rustls` doesn't have a license - not allowed
* `rustls-webpki` and `untrusted` has `ISC` which, in the current configuration, explicitly not allowed. Is it ok for MIT projects?

PR #183, opened 2024-01-19
Event listener interface and presence changes

parfeon:
feat(listener): change the real-time event handling interface

feat(presence-state): state maintained after set state is used

`user_id` state for specified channels will be maintained by the SDK. State with subscribe calls has been improved.

feat(api): adding first-class citizens to access subscription

Adding `Channel`, `ChannelGroup`, `ChannelMetadata` and `UuidMetadata` entities to be first-class citizens to access APIs related to them. Currently, access is provided only for subscription APIs.

feat(auto-retry): added ability to exclude endpoints from retry

Added ability to configure request retry policies to exclude specific endpoints from retry.


parfeon:
Adding PR for initial review purpose (interface shouldn't change) and documentation. There are still troubles with contract tests.

src/core/entity.rs
```diff
@@ -0,0 +1,196 @@
...
+};
+
+#[cfg(all(feature = "subscribe", feature = "std"))]
+use crate::{
+    core::{Deserializer, Transport},
+    lib::alloc::{sync::Arc, vec::Vec},
+    subscribe::{Subscribable, SubscribableType, Subscriber, Subscription, SubscriptionOptions},
+};
+
+pub(crate) trait PubNubEntity2 {
```
Xavrax, line 27: `PubNubEntity2` ?

src/core/retry_policy.rs
```diff
@@ -1,106 +1,354 @@
...
+            }
+            path if path.starts_with("/v1/message-actions/") => Endpoint::MessageReactions,
+            path if path.starts_with("/v1/channel-registration/") => Endpoint::ChannelGroups,
+            path if path.starts_with("/v2/objects/") => Endpoint::AppContext,
+            path if path.starts_with("/v1/push/") || path.starts_with("/v2/push/") => {
+                Endpoint::DevicePushNotifications
+            }
+            path if path.starts_with("/v1/files/") => Endpoint::Files,
+            _ => Endpoint::Unknown,
+        }
```
Xavrax, line 348: I like this syntax <3

src/dx/presence/builders/heartbeat.rs
```diff
@@ -339,4 +344,45 @@ mod it_should {
 
         assert!(matches!(result, Err(PubNubError::EffectCanceled)));
     }
+
+    // TODO: Make request cancelable
```
Xavrax, line 348: Do we want to release this with this PR?

src/dx/presence/builders/set_presence_state.rs
```diff
@@ -173,9 +193,16 @@ where
     /// Build and call asynchronous request.
     pub async fn execute(self) -> Result<SetStateResult, PubNubError> {
         let request = self.request()?;
+
+        #[cfg(feature = "std")]
+        if !request.channels.is_empty() {
+            request.on_execute.clone()(request.channels.clone(), request.state.clone());
```
Xavrax, line 199: Do we need to clone the PTR here? It seems like it is not moved out from this call.

src/dx/presence/event_engine/effects/mod.rs
```diff
@@ -184,10 +202,23 @@ impl Effect for PresenceEffect {
...
+                input,
+                executor,
+            } => {
+                heartbeat::execute(
+                    input,
+                    0,
+                    None,
+                    id,
+                    &RequestRetryConfiguration::None,
+                    executor,
```
Xavrax, line 216: Maybe not in this PR but it would be worth to prepare `HeartbeatExecuteParams` to make calling of `heartbeat::execute` more readable.

src/dx/subscribe/subscription.rs
```diff
@@ -0,0 +1,751 @@
...
+            }
+            false
+        });
+    }
+
+    fn id(&self) -> &String {
+        &self.id
+    }
+
+    fn client(&self) -> Weak<PubNubClientInstance<T, D>> {
```
Xavrax, line 505: Why do we return `Weak` and then `upgrade` it into strong pointer? ;o

src/dx/subscribe/subscription_set.rs
```diff
@@ -0,0 +1,953 @@
...
+    }
+}
+
+impl<T, D> EventSubscriber for SubscriptionSet<T, D>
+where
+    T: Transport + Send + Sync + 'static,
+    D: Deserializer + Send + Sync + 'static,
+{
+    fn subscribe(&self, cursor: Option<SubscriptionCursor>) {
+        log::debug!("~~~~~~~>>>> 3");
```
Xavrax, line 600: I know - WIP but note to not forget ;d

Xavrax (commented):
Great job!
First iteration of review here. 
Some questions and nitpicks.

Cargo.toml
```diff
@@ -106,13 +106,14 @@ getrandom = { version = "0.2", optional = true }
...
 
-# subscribe
+# subscribe, presence
 futures = { version = "0.3.28", default-features = false, optional = true }
 tokio = { version = "1", optional = true, features = ["rt-multi-thread", "macros", "time"] }
 async-channel = { version = "1.8", optional = true }
 
 # extra_platforms
 portable-atomic = { version = "1.3", optional = true, default-features = false, features = ["require-cas", "critical-section"] }
+backtrace = "0.3.69"
```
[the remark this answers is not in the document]
parfeon: Those are parts of debug. Thank you for reminding about it!

examples/presence_state.rs
```diff
@@ -1,16 +1,24 @@
+use std::collections::HashMap;
+
 use pubnub::{Keyset, PubNubClientBuilder};
-use serde::Serialize;
```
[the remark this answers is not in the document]
parfeon: Linter didn't liked it oO

examples/subscribe.rs
```diff
@@ -16,8 +23,8 @@ struct Message {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn snafu::Error>> {
-    let publish_key = env::var("SDK_PUB_KEY")?;
-    let subscribe_key = env::var("SDK_SUB_KEY")?;
+    let publish_key = "demo"; //env::var("SDK_PUB_KEY")?;
+    let subscribe_key = "demo"; //env::var("SDK_SUB_KEY")?;
```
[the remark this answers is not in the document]
parfeon: Yup :)

src/core/entity.rs
parfeon: This part is legacy. Need to tear it out.

src/dx/subscribe/subscription.rs
parfeon: We upgrade it only at the time when we need to use it. Rest of the time `Weak` pointer is stored which breaks retain cycle (subscriptions list stored in manager inside of `PubNubClientInstance`).

examples/presence_state.rs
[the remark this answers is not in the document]
Xavrax: Oooh, maybe they changed some of politics. 
Or maybe it was because State2 was not used during the check (just assuming - I don't know why it disagreed with that ;o)

src/dx/subscribe/subscription.rs
parfeon: Strange, my answer went somewhere. 
I'm trying to move away from strong references to the `PubNubClientInstance`. It will be great for it also return `Arc<Self>` when building in instance. When it will be able to pass components weak reference on itself, there won't be a way to create retain cycle which will require special approach to break (like we have right now with spawned EE). `Subscription` is stored in manager inside `PubNubClientInstance` - this is troublesome and the reason of weak reference.
`Subscription` itself already stores weak reference on it received from the entity for which it has been created. Entities so far store strong references.

examples/subscribe.rs
```diff
@@ -26,60 +33,122 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
...
+        .execute()
+        .await?;
+
+    tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
+
+    let subscription = client.subscription(
+        Some(&["my_channel", "other_channel"]),
+        None,
+        Some(vec![SubscriptionOptions::ReceivePresenceEvents]),
+    );
```
Xavrax, line 62: It seems to be a little bit unreadable. Can you provide at least the `SubscriptionParams` or something like that?

src/dx/subscribe/subscription.rs
```diff
@@ -0,0 +1,751 @@
...
+/// #     PubNubClientBuilder::with_reqwest_transport()
+/// #         .with_keyset(Keyset {
+/// #              subscribe_key: "demo",
+/// #              publish_key: Some("demo"),
+/// #              secret_key: Some("demo")
+/// #          })
+/// #         .with_user_id("uuid")
+/// #         .build()?;
+/// let channel = client.create_channel("my_channel");
+/// let subscription = channel.subscription(None);
```
Xavrax, line 57: This line is very confusing. Can you at least add comment what it does?

src/dx/subscribe/subscription.rs
```diff
@@ -0,0 +1,751 @@
...
+/// #     PubNubClientBuilder::with_reqwest_transport()
+/// #         .with_keyset(Keyset {
+/// #              subscribe_key: "demo",
+/// #              publish_key: Some("demo"),
+/// #              secret_key: Some("demo")
+/// #          })
+/// #         .with_user_id("uuid")
+/// #         .build()?;
+/// let channels = client.create_channels(&["my_channel_1", "my_channel_2"]);
+/// let subscription = channels[0].subscription(None).add(channels[1].subscription(None));
```
Xavrax, line 83: This chain is also very missleading. Can you explain what it exactly does?

examples/subscribe.rs
parfeon: But method itself provides documentation to the arguments? It is not enough? `SubscriptionParams` used by `EventEngine` already, and they are different from the set of arguments for this method.

src/dx/subscribe/subscription.rs
parfeon: Do you mean this line:
```rust
let subscription = channel.subscription(None);
```
?

Just leave a note that it will create `Subscription` instance for the entity?

src/dx/subscribe/subscription.rs
parfeon: I added those arithmetic operations before realizing that I'll probably need to work with `Arc<Subscription>` and `Arc<SubscriptionSet>`. Creating a separate struct with a single value just to add it - it will cause naming hell or not? 
I was planning to remove those traits implementation.

src/dx/subscribe/subscription.rs
Xavrax: Yeah exactly.

src/dx/subscribe/subscription.rs
Xavrax: I mean, I totally don't understand what means `channels[0].subscription(None).add(channels[1].subscription(None))`
It seems like an old C-style code that contains some macro values like
`do_something(value, NULL, NULL, NULL); `
I can't tell what should be better in this case, but for random reader (e.g. @techwritermat ) it can be a little bit hard to understand what is happening here.

examples/subscribe.rs
Xavrax: It's more about convention. 
I prefer when you exactly know what do you try to achieve. 
That's why I propose `SubscriptionParams` or just a builder for that. 

It will explicit explain what options you set here.

src/core/channel.rs
```diff
@@ -0,0 +1,217 @@
...
+#[derive(Debug)]
+pub struct ChannelRef<T, D> {
+    /// Reference on backing [`PubNubClientInstance`] client.
+    ///
+    /// Client is used to support entity-specific actions like:
+    /// * subscription
+    ///
+    /// [`PubNubClientInstance`]: PubNubClientInstance
+    #[allow(unused)]
+    client: Arc<PubNubClientInstance<T, D>>,
```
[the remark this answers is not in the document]
parfeon: Nah, it appears to be used in tests (`subscribe` feature use it, but if I'll hide it behind condition then `T`, and `D` will be unused and clippy will smack me again :/ )
So I'm keeping this for now.

src/core/entity.rs
parfeon: Removed

src/dx/presence/builders/heartbeat.rs
parfeon: Probably next one will be fine :)

src/dx/presence/event_engine/effects/mod.rs
parfeon: Will do in next PRs.

src/dx/subscribe/subscription_set.rs
parfeon: Cleaned.

examples/subscribe.rs
parfeon: Added `SubscriptionParams`

src/dx/subscribe/subscription.rs
parfeon: I've made proper implementation, now there is simple additions :)

parfeon:
@pubnub-release-bot release

techwritermat (approved):

PR #184, opened 2024-01-31
Fix `cursor` reset issue and cancelled `effects` spawning

parfeon:
feat(subscription): add more types to create `SubscriptionCursor` from

Make it possible to create `SubscriptionCursor` from the string slice.

feat(subscription): add methods to manipulate subscriptions in set

Add `add_subscriptions(..)` and `sub_subscriptions(..)` to `SubscriptionSet` to make it possible in addition to sets manipulation use list of subscriptions.

fix(subscribe): fix `cursor` reset issue
    
Fix issue because of which `cursor` is not reset on `Subscription` and `SubscriptionSet` on unsubscribe.

fix(subscription): fix issue with cancelled effects

Fix issue because of which cancelled effects still asynchronously spawned for processing.

docs(inline): change `client` to `pubnub` in inline docs

refactor(subscribe): add subscription token validation
 
Added a method to validate the provided subscription token to conform to PubNub time token requirements with precision.
    
refactor(examples): refactor `subscribe` example
    
Separate `subscribe` example into two to show separately `subscribe` feature and `presence state` maintenance with subscribe.

### DIFF
--- a/examples/subscribe.rs
+++ b/examples/subscribe.rs
@@ -45,7 +45,7 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
         channel_groups: None,
         options: Some(vec![SubscriptionOptions::ReceivePresenceEvents]),
     });
-    subscription.subscribe(None);
+    subscription.subscribe();
     let subscription_clone = subscription.clone_empty();
 
     // Attach connection status to the PubNub client instance.
@@ -130,7 +130,9 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
     // drop(subscription_clone);
 
     println!(
-        "\nUnsubscribe from all data streams. To restore requires `subscription.subscribe(None)` call." );
+        "\nUnsubscribe from all data streams. To restore call `subscription.subscribe()` or \
+        `subscription.subscribe_with_timetoken(Some(<timetoken>)) call."
+    );
     // Clean up before complete work with PubNub client instance.
     pubnub.unsubscribe_all();
     tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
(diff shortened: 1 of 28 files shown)


examples/subscribe.rs
```diff
@@ -40,36 +38,25 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
...
     tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
 
-    let subscription = client.subscription(SubscriptionParams {
+    let subscription = pubnub.subscription(SubscriptionParams {
         channels: Some(&["my_channel", "other_channel"]),
         channel_groups: None,
         options: Some(vec![SubscriptionOptions::ReceivePresenceEvents]),
     });
-    subscription.subscribe(None);
+    subscription.subscribe();
```
Xavrax, line 48: I love that getting rid of this meaningless `None` in `subscribe()`. 

But i'm still thinking about changing the `SubscriptionParams` into builder. What do you think?

examples/subscribe.rs
```diff
@@ -147,9 +130,11 @@ async fn main() -> Result<(), Box<dyn snafu::Error>> {
     // drop(subscription_clone);
```
Xavrax, line 131: Very useful comments <3
That's very important that we unsubscribe on drop.

Xavrax (approved):
LGTM! Great work!

Only one proposal but it can be done in other PR (if we want to).

examples/subscribe.rs
parfeon: We do actually :)

examples/subscribe.rs
parfeon: Maybe in next changes because @techwritermat almost done with docs.

parfeon:
@pubnub-release-bot release

PR #195, opened 2026-06-05
Enable HTTP/2 via default Reqwest transport and log negotiated protocol

KGronek-Pubnub:
feat: Enable HTTP/2 negotiation on the default `TransportReqwest` by turning on reqwest’s `native-tls-alpn` feature, so ALPN can select `h2` when the origin supports it while preserving HTTP/1.1 fallback.

refactor:  Add debug-level protocol logging in async and blocking `TransportReqwest::send()` after each completed request, including HTTP method, path, and the Reqwest-reported protocol version.

fix: Fix a Clippy `derivable_impls` lint in `RequestRetryConfiguration` (derive `Default` on the enum instead of a manual impl).


parfeon (approved):
LGTM!

KGronek-Pubnub:
@pubnub-release-bot release

